PR #2, opened 2016-04-12
expose process module to renderer "processes"

mykmelez:
This branch exposes the `process` module to renderer "processes" by defining a WebIDL interface implemented by an XPCOM component that proxies access to the `process` module.

The branch enables the "hello world" app to display version numbers for Node/Chromium/Electron, although the version numbers themselves are for no-Node (a hardcoded `0`), Gecko (`48.0a1`), and Positron (also `48.0a1`).

Note that it branches off from the multiple-module-loaders branch in #1, so it includes those changes. To see just the changes on this branch, visit:

https://github.com/mykmelez/positron/compare/multiple-module-loaders...renderer-process-modules


positron/components/ModuleLoader.jsm
```diff
@@ -42,28 +42,32 @@ const globalPaths = [
...
-function Require(requirer) {
+
+function ModuleLoader(processType) {
+  /**
+   * Mapping from module IDs (resource: URLs) to module objects.
+   *
+   * @keys {string} The ID (resource: URL) for the module.
+   * @values {object} An object representing the module.
+   */
+  let modules = new Map();
```
marcoscaceres: Nit: const

mykmelez: > Nit: const

Fixed in ac43d30.

positron/components/Process.manifest
```diff
@@ -0,0 +1,3 @@
+component {3c81d709-5fb4-4144-9612-9ecc1be4e7b1} Process.js
+contract @mozilla.org/positron/process;1 {3c81d709-5fb4-4144-9612-9ecc1be4e7b1}
+category JavaScript-global-property process @mozilla.org/positron/process;1
```
khuey: Hmm, I'm not sure this is the best way to do this long term.  I think DOM wants to get rid of things like this.  But it's ok for now.

PR #11, opened 2016-04-12
implement app.quit() (fixes #9)

### LINKED ISSUE #9: implement app.quit
The "hello world" app calls *app.quit* to quit the app when all windows have closed (unless Mac), so we should implement it.

mykmelez:


positron/modules/atom_browser_app.js
```diff
@@ -4,7 +4,20 @@
 
 "use strict";
 
-exports.app = { /* stub */ };
+const Cc = Components.classes;
+const Ci = Components.interfaces;
+const Cr = Components.results;
+const Cu = Components.utils;
+
```
marcoscaceres: maybe: 
```JS
const { classes: Cc, interfaces: Ci, results: Cr, utils: Cu } = Components;
```
Also, are all those used?

positron/modules/atom_browser_app.js
```diff
@@ -4,7 +4,20 @@
...
+Cu.import('resource://gre/modules/Services.jsm');
+
+exports.app = {
+  quit: function() {
```
marcoscaceres: You don't need function(), it's implied. I.e., just `quit(){ ... },`

positron/modules/atom_browser_app.js
mykmelez: > const { classes: Cc, interfaces: Ci, results: Cr, utils: Cu } = Components;

Indeed, that's less verbose. Fixed in b3f8353.

mykmelez: > Also, are all those used?

They aren't, but I'd prefer to make this boilerplate consistent across modules that access Components.

positron/modules/atom_browser_app.js
mykmelez: > You don't need function(), it's implied. I.e., just quit(){ ... },

Ah, method definitions, nifty. Fixed in 3d95433.

PR #14, opened 2016-04-13
import Electron tests

mykmelez:
Besides importing the Electron tests (which are simply an Electron app), this makes PositronAppRunner accommodate *main* scripts that are paths and expands the README with docs on running apps and tests.


README.md
```diff
@@ -19,3 +19,26 @@ Build Command:
...
+
+You can also `npm link` the build directory and then run the app via the *positron* command:
+
+```bash
+(cd obj-x86_64-apple-darwin14.5.0/dist/; npm link) # Build dir name will vary.
+positron positron/test/hello-world
+```
+
+### How to test
+To start an Electron test run:
```
marcoscaceres: Nit: Positron :wink:

marcoscaceres: oh, actually, NM... I guess these are Electron tests.

positron/components/PositronAppRunner.jsm
```diff
@@ -53,14 +54,14 @@ PositronAppRunner.prototype = {
...
-    mainScript.append(aScriptName);
+  _loadMainScript: function par_LoadMainScript(aScriptPath) {
+    let mainScriptPath = OS.Path.join(this._baseDir.path, aScriptPath);
+    let mainScriptURI = OS.Path.toFileURI(mainScriptPath);
     let requirer = {
       id: "resource:///modules/PositronAppRunner.jsm",
       exports: {},
     };
-    (new Require(requirer))(Services.io.newFileURI(mainScript).spec);
+    (new Require(requirer))(mainScriptURI);
```
marcoscaceres: Not something to fix here... but Require() is an object that instantiates to a function? That seems very odd O_o. Is it too late to fix this (i.e., should be `const require = makeRequire(requirer)` and not a constructor, or am I missing something)?

mykmelez: > Not something to fix here... but Require() is an object that instantiates to a function?

Yes, Require is a constructor that returns a function (which isn't instanceof Require).

> That seems very odd O_o.

Indeed, I'm not thrilled with it either.

> Is it too late to fix this (i.e., should be const require = makeRequire(requirer) and not a constructor, or am I missing something)?

It isn't too late to fix anything in Positron!

Note that the module loader is a stopgap intended to enable us to make progress on Electron integration until we integrate SpiderNode, at which point I expect that we'll switch to Node's module loader. Nevertheless, it should still expose a sensible interface.

In this case, I've already modified the interface over in #1, such that this code will become:

```
let loader = new ModuleLoader("browser" /* processType */);
loader.require(requirer, mainScriptURI);
```

Once both that PR and this one are merged.

And yes, we can also s/let/const/, although in this case we don't actually need to persist a reference to the module loader, so it can be even more simply:

```
(new ModuleLoader("browser")).require(requirer, mainScriptURI);
```

mykmelez: > And yes, we can also s/let/const/, although in this case we don't actually need to persist a reference to the module loader, so it can be even more simply:
> 
> (new ModuleLoader("browser")).require(requirer, mainScriptURI);

Fixed in conjunction with a merge from master.

mykmelez:
@marcoscaceres Is there anything else you'd like to see here?

marcoscaceres:
Nope :)

PR #15, opened 2016-04-15
Create basic branding structure

jryans:
As part of working on DevTools support, I noticed there's no branding files in place yet. Some platform modules expect to be able to load files like "brand.properties" without catching errors, so it seems reasonable to set up these files.

@mykmelez, maybe you can review?


positron/branding/unofficial/locales/moz.build
```diff
@@ -0,0 +1,7 @@
+# -*- Mode: python; c-basic-offset: 4; indent-tabs-mode: nil; tab-width: 40 -*-
+# vim: set filetype=python:
+# This Source Code Form is subject to the terms of the Mozilla Public
+# License, v. 2.0. If a copy of the MPL was not distributed with this
+# file, You can obtain one at http://mozilla.org/MPL/2.0/.
+
+JAR_MANIFESTS += ['jar.mn']
```
mykmelez: Nit: trailing newline.

positron/branding/unofficial/locales/en-US/brand.dtd
```diff
@@ -0,0 +1,8 @@
+<!-- This Source Code Form is subject to the terms of the Mozilla Public
+   - License, v. 2.0. If a copy of the MPL was not distributed with this
+   - file, You can obtain one at http://mozilla.org/MPL/2.0/. -->
+
+<!ENTITY  brandShortName  "Positron">
+<!ENTITY  brandFullName   "Mozilla Positron">
+<!ENTITY  vendorShortName "Mozilla">
+<!ENTITY  logoTrademark   "">
```
mykmelez: It isn't clear where this'll be used, although it shouldn't cause any problem to include it.

jryans: Yeah, I suppose this one could be removed... It's only shown on the about popup it seems, so it's not being read from code. It's probably clearer to not define it then.

jryans:
Okay, all comments addressed @mykmelez.

mykmelez:
Thanks, @jryans!

PR #16, opened 2016-04-15
First pass at DevTools for BrowserWindow

### LINKED ISSUE #3: open Gecko devtools on BrowserWindow.webContents.openDevTools()
The "hello world" app invokes `BrowserWindow.webContents.openDevTools()` to open a Chromium devtools frame in its main window in Electron. Positron should do the same with Gecko devtools.

jryans:
I've added support to open the DevTools when calling `BrowserWindow.webContents.openDevTools()` as in the example app (so this fixes #3).

I added some additional abilities to the tools for targeting specific windows directly to help with this case. I've kept all the `/devtools` changes in separate commits from `/positron` ones, for hopefully easier merging if bits are landed in m-c later.

~~The tools seems to be up and running in general, but there's a few rough edges for now, like you get a giant error on quit if the tools are still open. I figured I would resolve these things separately.~~ (Error on quit now resolved.)

@mykmelez, maybe you can review?

### DIFF
--- a/devtools/client/framework/target-from-url.js
+++ b/devtools/client/framework/target-from-url.js
@@ -4,8 +4,6 @@
 
 "use strict";
 
-const { Cu, Ci } = require("chrome");
-
 const { TargetFactory } = require("devtools/client/framework/target");
 const { DebuggerServer } = require("devtools/server/main");
 const { DebuggerClient } = require("devtools/shared/client/main");
@@ -14,27 +12,32 @@ const { Task } = require("resource://gre/modules/Task.jsm");
 /**
  * Construct a Target for a given URL object having various query parameters:
  *
- * type: tab, process
- *    {String} The type of target to connect to.  Currently tabs and processes are supported types.
+ * type: tab, process, window
+ *    {String} The type of target to connect to.
  *
- * If type="tab":
+ * If type == "tab":
  * id:
  *    {Number} the tab outerWindowID
  * chrome: Optional
- *    {Boolean} Force the creation of a chrome target. Gives more privileges to the tab
- *    actor. Allows chrome execution in the webconsole and see chrome files in
- *    the debugger. (handy when contributing to firefox)
+ *    {Boolean} Force the creation of a chrome target. Gives more privileges to
+ *    the tab actor. Allows chrome execution in the webconsole and see chrome
+ *    files in the debugger. (handy when contributing to firefox)
  *
- * If type="process":
+ * If type == "process":
  * id:
- *    {Number} the process id to debug. Default to 0, which is the parent process.
+ *    {Number} the process id to debug. Default to 0, which is the parent
+ *    process.
+ *
+ * If type == "window":
+ * id:
+ *    {Number} the window outerWindowID
  *
  * @param {URL} url
  *        The url to fetch query params from.
  *
  * @return A target object
  */
-exports.targetFromURL = Task.async(function*(url) {
+exports.targetFromURL = Task.async(function* (url) {
   let params = url.searchParams;
   let type = params.get("type");
   if (!type) {
@@ -85,6 +88,26 @@ exports.targetFromURL = Task.async(function*(url) {
     } catch(ex) {
       if (ex.error == "noProcess") {
         throw new Error("targetFromURL, process with id:'" + id+ "' doesn't exist");
+      }
+      throw ex;
+    }
+  } else if (type == "window") {
+    // Fetch target for a remote window actor
+    DebuggerServer.allowChromeProcess = true;
+    try {
+      id = parseInt(id, 10);
+      if (isNaN(id)) {
+        throw new Error("targetFromURL, window requires id parameter");
+      }
+      let response = yield client.mainRoot.getWindow({
+        outerWindowID: id,
+      });
+      form = response.window;
+      chrome = true;
+    } catch (ex) {
+      if (ex.error == "notFound") {
+        throw new Error(`targetFromURL, window with id:'${id}' ` +
+                        "doesn't exist");
       }
       throw ex;
     }
--- a/devtools/client/framework/toolbox-init.js
+++ b/devtools/client/framework/toolbox-init.js
@@ -26,6 +26,19 @@ if (url.search.length > 1) {
                    .getInterface(Ci.nsIDOMWindowUtils)
                    .containerElement;
 
+  // If there's no containerElement, use the current window.
+  if (!host) {
+    host = {
+      contentWindow: window,
+      contentDocument: document,
+      // toolbox.js wants to set attributes on the frame that contains it, but
+      // that is fine to skip and doesn't make sense when using the current
+      // window.
+      setAttribute() {},
+      ownerDocument: document,
+    };
+  }
+
   // Specify the default tool to open
   let tool = url.searchParams.get("tool");
 
--- a/devtools/client/framework/toolbox-init.js
+++ b/devtools/client/framework/toolbox-init.js
@@ -42,6 +42,7 @@ if (url.search.length > 1) {
   // Specify the default tool to open
   let tool = url.searchParams.get("tool");
 
+  let toolboxOpened;
   if (url.searchParams.has("target")) {
     // Attach toolbox to a given browser iframe (<xul:browser> or <html:iframe
     // mozbrowser>) whose reference is set on the host iframe.
@@ -64,22 +65,40 @@ if (url.search.length > 1) {
         DebuggerServer.addBrowserActors();
       }
       let client = new DebuggerClient(DebuggerServer.connectPipe());
-      Task.spawn(function*() {
+      toolboxOpened = Task.spawn(function*() {
         yield client.connect();
         // Creates a target for a given browser iframe.
         let response = yield client.getTab({ tab });
         let form = response.tab;
-        let target = yield TargetFactory.forRemoteTab({client, form, chrome: false});
+        let target = yield TargetFactory.forRemoteTab({
+          client,
+          form,
+          chrome: false
+        });
         let options = { customIframe: host };
-        yield gDevTools.showToolbox(target, tool, Toolbox.HostType.CUSTOM, options);
+        return gDevTools.showToolbox(target, tool, Toolbox.HostType.CUSTOM,
+                                     options);
       });
     }
   } else {
-    targetFromURL(url).then(target => {
+    toolboxOpened = Task.spawn(function*() {
+      let target = yield targetFromURL(url);
       let options = { customIframe: host };
-      return gDevTools.showToolbox(target, tool, Toolbox.HostType.CUSTOM, options);
-    }).then(null, e => {
-      window.alert("Unable to start the toolbox:" + e.message);
+      return gDevTools.showToolbox(target, tool, Toolbox.HostType.CUSTOM,
+                                   options);
     });
   }
+
+  toolboxOpened.catch(e => {
+    window.alert("Unable to start the toolbox: " + e.message);
+  });
+
+  // Ensure the toolbox gets destroyed if the host unloads
+  toolboxOpened.then(toolbox => {
+    let hostUnload = () => {
+      host.contentWindow.removeEventListener("unload", hostUnload);
+      toolbox.destroy();
+    };
+    host.contentWindow.addEventListener("unload", hostUnload);
+  });
 }
--- a/devtools/server/actors/root.js
+++ b/devtools/server/actors/root.js
@@ -14,6 +14,8 @@ const { DebuggerServer } = require("devtools/server/main");
 loader.lazyGetter(this, "ppmm", () => {
   return Cc["@mozilla.org/parentprocessmessagemanager;1"].getService(Ci.nsIMessageBroadcaster);
 });
+loader.lazyRequireGetter(this, "WindowActor",
+  "devtools/server/actors/window", true);
 
 /* Root actor for the remote debugging protocol. */
 
@@ -163,12 +165,15 @@ RootActor.prototype = {
     // Added in Firefox 40. Indicates that the backend supports registering custom
     // commands through the WebConsoleCommands API.
     webConsoleCommands: true,
-    // Whether root actor exposes tab actors
-    // if allowChromeProcess is true, you can fetch a ChromeActor instance
-    // to debug chrome and any non-content ressource via getProcess request
-    // if allocChromeProcess is defined, but not true, it means that root actor
-    // no longer expose tab actors, but also that getProcess forbids
-    // exposing actors for security reasons
+    // Whether root actor exposes tab actors and access to any window.
+    // If allowChromeProcess is true, you can:
+    // * get a ChromeActor instance to debug chrome and any non-content
+    //   resource via getProcess requests
+    // * get a WindowActor instance to debug windows which could be chrome,
+    //   like browser windows via getWindow requests
+    // If allowChromeProcess is defined, but not true, it means that root actor
+    // no longer expose tab actors, but also that the above requests are
+    // forbidden for security reasons.
     get allowChromeProcess() {
       return DebuggerServer.allowChromeProcess;
     },
@@ -220,6 +225,7 @@ RootActor.prototype = {
     this.conn = null;
     this._tabActorPool = null;
     this._globalActorPool = null;
+    this._windowActorPool = null;
     this._parameters = null;
     this._chromeActor = null;
   },
@@ -322,6 +328,38 @@ RootActor.prototype = {
                  message: "Unexpected error while calling getTab(): " + error };
       }
     });
+  },
+
+  onGetWindow: function ({ outerWindowID }) {
+    if (!DebuggerServer.allowChromeProcess) {
+      return {
+        from: this.actorID,
+        error: "forbidden",
+        message: "You are not allowed to debug windows."
+      };
+    }
+    let window = Services.wm.getOuterWindowWithId(outerWindowID);
+    if (!window) {
+      return {
+        from: this.actorID,
+        error: "notFound",
+        message: `No window found with outerWindowID ${outerWindowID}`,
+      };
+    }
+
+    if (!this._windowActorPool) {
+      this._windowActorPool = new ActorPool(this.conn);
+      this.conn.addActorPool(this._windowActorPool);
+    }
+
+    let actor = new WindowActor(this.conn, window);
+    actor.parentID = this.actorID;
+    this._windowActorPool.addActor(actor);
+
+    return {
+      from: this.actorID,
+      window: actor.form(),
+    };
   },
 
   onTabListChanged: function () {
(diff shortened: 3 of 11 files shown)


bgrins:
Is there a corresponding bug to merge this work back into m-c?

bgrins:
> Is there a corresponding bug to merge this work back into m-c?

Or maybe it doesn't make sense to do that at this point.  I was thinking we'd bump into conflicts trying to keep it in sync but these files don't seem too likely to drift.

jryans:
No bug to merge to m-c yet. So far the devtools changes are pretty isloated, so I think it should be okay.

mykmelez:
I can review, but it'd be helpful to have someone more familiar with the /devtools code look at that part of it. Perhaps @ochameau?

devtools/client/framework/toolbox-init.js
```diff
@@ -51,22 +62,32 @@ if (url.search.length > 1) {
...
-    targetFromURL(url).then(target => {
+    toolboxOpened = Task.spawn(function*() {
+      let target = yield targetFromURL(url);
       let options = { customIframe: host };
       return gDevTools.showToolbox(target, tool, Toolbox.HostType.CUSTOM, options);
-    }).then(null, e => {
-      window.alert("Unable to start the toolbox:" + e.message);
+    }).catch(e => {
+      window.alert("Unable to start the toolbox: " + e.message);
     });
```
mykmelez: This *catch* call should rethrow the exception, otherwise this *toolboxOpened* promise will be resolved instead of rejected. Alternately, move the *catch* call to the `toolboxOpened.then()` chain below, so it catches and reports exceptions in both *toolboxOpened* promises.

devtools/server/docs/actor-hierarchy.md
```diff
@@ -63,6 +63,11 @@ and returns its `actorID`. That's the main role of RootActor.
    |   Returned by "listWorkers" request to a ChildProcessActor to get workers
    |   for the chrome of the child process.
    |
+   |-- WindowActor (window.js)
+   |   Targets a single window, such as a browser window in Firefox, but it can
+   |   be used to reach any window in the process.
+   |   Returned by "getWindow" request to the root actor.
+   |
```
mykmelez: Hmm, this tree doesn't render well, not by GitHub's MarkDown renderer anyway:

https://github.com/jryans/positron/blob/browser-window-devtools/devtools/server/docs/actor-hierarchy.md

That isn't caused by your change! So can be addressed separately. But it's an easy fix, you might consider rolling it into this branch. I submitted a PR for it:

https://github.com/jryans/positron/pull/1

positron/modules/atom_browser_web_contents.js
```diff
@@ -29,6 +32,57 @@ let WebContents_prototype = {
...
   },
+
+  openDevTools: function() {
+    // TODO: When tools can be opened inside the content window, support
+    // `detach` option to force into a new window instead.
+
+    // Ensure DevTools core modules are loaded, including support for the about
+    // URL below which is registered dynamically.
+    const { require } = Cu.import("resource://devtools/shared/Loader.jsm", {});
+    require("devtools/client/main");
```
mykmelez: This will mask the global *require*, but that should be ok, since it'll only do so for this function, which doesn't need the global *require*.

positron/modules/atom_browser_web_contents.js
```diff
@@ -29,6 +32,57 @@ let WebContents_prototype = {
...
+    }
+    let onMessage = ({ data }) => {
+      data = JSON.parse(data);
+      if (data.name !== "toolbox-close") {
+        return;
+      }
+      toolsWindow.close();
+    };
+    // TODO: Improve browser window close handling.  DevTools is currently still
+    // attached when the window closes, leading to large errors.
```
mykmelez: What about listening for *beforeunload* instead? Alternately, you could try listening for *pagehide* or registering a *domwindowclosed* observer via *nsIWindowWatcher.registerNotification*, although I suspect those events/notifications won't arrive in time either.

positron/modules/atom_browser_web_contents.js
```diff
@@ -29,6 +32,57 @@ let WebContents_prototype = {
   loadURL: function(url) {
     this._browserWindow._domWindow.location = url;
   },
+
+  openDevTools: function() {
```
mykmelez: Over in #11, @marcoscaceres reminded me about method definitions, which would make this definition simpler.

devtools/client/framework/toolbox-init.js
```diff
@@ -51,22 +62,32 @@ if (url.search.length > 1) {
...
   }
+
+  // Ensure the toolbox gets destroyed if the host unloads
+  toolboxOpened.then(toolbox => {
+    let hostUnload = () => {
+      host.contentWindow.removeEventListener("unload", hostUnload);
+      toolbox.destroy();
+    };
+    host.contentWindow.addEventListener("unload", hostUnload);
+  });
```
ochameau: Doesn't the toolbox iframe get unloaded?
i.e. don't we already listen for toolbox iframe unload?
It would look simplier if we listen on `window` rather than `host.contentWindow`.

devtools/server/actors/window.js
```diff
@@ -0,0 +1,40 @@
...
+ * Creates a WindowActor for debugging a single window, like a browser window in
+ * Firefox, but it can be used to reach any window in the process.  Most of the
+ * implementation is inherited from TabActor. WindowActor exposes all tab actors
+ * via its form() request, like TabActor.
+ *
+ * You can request a specific window's actor via RootActor.getWindow().
+ *
+ * @param connection DebuggerServerConnection
+ *        The connection to the client.
+ * @param window mozIDOMWindowProxy
```
ochameau: I would just have said DOMWindow instead of mozIDOMWindowProxy.

devtools/server/actors/window.js
```diff
@@ -0,0 +1,40 @@
...
+  Object.defineProperty(this, "docShell", {
+    value: docShell,
+    configurable: true
+  });
+}
+
+WindowActor.prototype = Object.create(TabActor.prototype);
+
+// TODO: This setting is mysteriously named, we should split up the
+// functionality that is triggered by it.
```
ochameau: Didn't we opened a bug about that yet?

devtools/server/docs/actor-hierarchy.md
```diff
@@ -63,6 +63,11 @@ and returns its `actorID`. That's the main role of RootActor.
    |   Returned by "listWorkers" request to a ChildProcessActor to get workers
    |   for the chrome of the child process.
    |
+   |-- WindowActor (window.js)
+   |   Targets a single window, such as a browser window in Firefox, but it can
+   |   be used to reach any window in the process.
```
ochameau: s/the process/the parent process/

positron/modules/atom_browser_web_contents.js
```diff
@@ -29,6 +32,57 @@ let WebContents_prototype = {
...
+    let url = `about:devtools-toolbox?type=window&id=${id}`;
+    let features = "chrome,resizable,centerscreen," +
+                   "width=1024,height=768";
+    let toolsWindow = Services.ww.openWindow(null, url, null, features, null);
+
+    let onLoad = () => {
+      toolsWindow.removeEventListener("load", onLoad);
+      toolsWindow.addEventListener("unload", onUnload);
+      window.addEventListener("unload", onBrowserUnload);
+      this.emit("devtools-opened");
```
ochameau: These events don't seem to be used anywhere?

positron/modules/atom_browser_web_contents.js
```diff
@@ -29,6 +32,57 @@ let WebContents_prototype = {
...
+      toolsWindow.close();
+    };
+    // TODO: Improve browser window close handling.  DevTools is currently still
+    // attached when the window closes, leading to large errors.
+    let onBrowserUnload = () => {
+      toolsWindow.close();
+    };
+
+    toolsWindow.addEventListener("message", onMessage);
+    toolsWindow.addEventListener("load", onLoad);
```
ochameau: Is it any useful to listent for toolsWindow's load?
Couldn't we listen for toolsWindow.unload and window.unload immediately?

positron/modules/atom_browser_web_contents.js
ochameau: I do not see errors? It looks like, here, you have a lot of code to handle closing. Isn't that enough?? Or it is just some races?

positron/modules/atom_browser_web_contents.js
ochameau: You could use `const { loader } = Cu.import(); loader.require();` to mitigate that.

ochameau:
I think we could reasonably land the devtools patch with a new test for getWindow in devtools/client/framework/test/browser_target_from_url.js
Regarding unloading issues. I would really like to make it so that regular toolboxes used about: url. And remove the direct calls of Toolbox.xul from chrome, so that we kills some chrome interactions for the benefit of devtools.html. I imagine I would hit same issues than you and I'm easer to know more about unload issues.

positron/modules/atom_browser_web_contents.js
jryans: On OS X at least for me, when I quit Positron (Cmd-Q) with both app window and toolbox open, the DevTools actors are still bound to the app window, and they get a window destroyed notification, but then try to keep using the window which no longer exists, and explode in various ways.

positron/modules/atom_browser_web_contents.js
jryans: Yep, I can switch! I use those for new files but here I was matching the style of the other methods. Anyway, I can change.

devtools/server/actors/window.js
jryans: A search for `isRootActor` did not reveal an existing bug to me, but I can file one.

positron/modules/atom_browser_web_contents.js
jryans: That's correct, they are part of the Electron API for DevTools on [`BrowserWindow.webContents`][1]. We don't have any clear test plan at the moment, so I did not add any usage code myself.

[1]: http://electron.atom.io/docs/v0.37.5/api/web-contents/#event-devtools-opened

positron/modules/atom_browser_web_contents.js
jryans: When I did that at first, I got an initial `unload` event from the tool window, possibly from moving from `about:blank` to the actual tool window page.

positron/modules/atom_browser_web_contents.js
jryans: I'll likely change to `loader` explicitly to reduce confusion.

devtools/server/docs/actor-hierarchy.md
jryans: Thanks, I'll roll it in!

positron/modules/atom_browser_web_contents.js
mykmelez: Indeed, we aren't running automated tests yet, but the plan is to start running the Electron tests that are landing in #14. That PR just lands the tests, though. Actually running them will take more work.

devtools/client/framework/toolbox-init.js
jryans: @ochameau noted on IRC that we should move this into `toolbox.js` to cover all toolboxes.

devtools/client/framework/target-from-url.js
```diff
@@ -88,6 +91,27 @@ exports.targetFromURL = Task.async(function*(url) {
       }
       throw ex;
     }
+  } else if (type == "window") {
+    // Fetch target for a remote window actor
+    DebuggerServer.allowChromeProcess = true;
+    try {
+      id = parseInt(id, 10);
+      if (isNaN(id)) {
+        throw new Error("targetFromURL, " +
```
marcoscaceres: nit: this all fits on one line? Also, can we please avoid doing white space like this (as it makes it hard to use beautifiers)? Just use 2 spaces everywhere.

devtools/client/framework/toolbox-init.js
```diff
@@ -26,9 +26,20 @@ if (url.search.length > 1) {
                    .getInterface(Ci.nsIDOMWindowUtils)
                    .containerElement;
 
+  // If there's no containerElement, use the current window.
+  if (!host) {
+    host = {
+      contentWindow: window,
+      contentDocument: document,
+      setAttribute() {},
```
marcoscaceres: Maybe add a comment explaining why this is a noop?

devtools/client/framework/target-from-url.js
```diff
@@ -71,7 +74,7 @@ exports.targetFromURL = Task.async(function*(url) {
     // Fetch target for a remote chrome actor
     DebuggerServer.allowChromeProcess = true;
     try {
-      id = parseInt(id);
+      id = parseInt(id, 10);
```
marcoscaceres: Should we fix the parseInt on line 60 too?

marcoscaceres: There is a lot of redundancy wrt the id in this code and below. Why not deal with it on line 46, where id is first set up? At least there, it could be parseInt()'d to get either the NaN or number. Then we wouldn't need to do this check after each `if`. It might help flatten the code a bit.

Also, because "type" is working as a enum, this code is kinda screaming to be made into a switch block.

devtools/client/framework/toolbox-init.js
```diff
@@ -51,22 +62,32 @@ if (url.search.length > 1) {
...
-    }).then(null, e => {
-      window.alert("Unable to start the toolbox:" + e.message);
+    }).catch(e => {
+      window.alert("Unable to start the toolbox: " + e.message);
     });
   }
+
+  // Ensure the toolbox gets destroyed if the host unloads
+  toolboxOpened.then(toolbox => {
+    let hostUnload = () => {
```
marcoscaceres: nit: const

devtools/server/actors/root.js
```diff
@@ -14,6 +14,8 @@ const { DebuggerServer } = require("devtools/server/main");
 loader.lazyGetter(this, "ppmm", () => {
   return Cc["@mozilla.org/parentprocessmessagemanager;1"].getService(Ci.nsIMessageBroadcaster);
 });
+loader.lazyRequireGetter(this, "WindowActor",
+  "devtools/server/actors/window", true);
```
marcoscaceres: Nit: please add a comment to explain the "true"

devtools/server/actors/root.js
```diff
@@ -163,12 +165,15 @@ RootActor.prototype = {
     // Added in Firefox 40. Indicates that the backend supports registering custom
     // commands through the WebConsoleCommands API.
     webConsoleCommands: true,
-    // Whether root actor exposes tab actors
-    // if allowChromeProcess is true, you can fetch a ChromeActor instance
-    // to debug chrome and any non-content ressource via getProcess request
-    // if allocChromeProcess is defined, but not true, it means that root actor
-    // no longer expose tab actors, but also that getProcess forbids
-    // exposing actors for security reasons
+    // Whether root actor exposes tab actors and access to any window
```
marcoscaceres: Nit: full stop at the end of this line?

devtools/server/actors/root.js
```diff
@@ -324,6 +330,38 @@ RootActor.prototype = {
...
+  onGetWindow: function ({ outerWindowID }) {
+    if (!DebuggerServer.allowChromeProcess) {
+      return {
+        from: this.actorID,
+        error: "forbidden",
+        message: "You are not allowed to debug windows."
+      };
+    }
+    let window = Services.wm.getOuterWindowWithId(outerWindowID);
+    if (!window) {
```
marcoscaceres: Nit: maybe just `if (!Services.wm.getOuterWindowWithId(outerWindowID))`. Otherwise, change `let` to `const` to signal it won't change.

marcoscaceres: Ah, I see you use window below, should still be `const` tho.

devtools/shared/client/main.js
```diff
@@ -1654,6 +1654,27 @@ RootClient.prototype = {
...
+   *
+   * @param number outerWindowID
+   *        The outerWindowID of the top level window you are looking for.
+   */
+  getWindow: function({ outerWindowID }) {
+    if (!outerWindowID) {
+      throw new Error("Must specify outerWindowID");
+    }
+
+    let packet = {
```
marcoscaceres: nit const

positron/modules/atom_browser_web_contents.js
jryans: After talking to @ochameau on IRC, I realized the new `WindowActor` wasn't watching for chrome docshells. With that fixed, the tools shutdown correctly, so I was able to remove this particular listener.

devtools/client/framework/toolbox-init.js
jryans: Thanks, I moved the `catch` portion below so catches both flows.

devtools/server/actors/window.js
jryans: Okay, changed.

devtools/client/framework/toolbox-init.js
jryans: It's unclear to me if this matters for all toolboxes, or just the specific path being used here. I think I'll leave moving this into `toolbox.js` as a separate task.

devtools/server/actors/window.js
jryans: Filed https://bugzilla.mozilla.org/show_bug.cgi?id=1266561 for this and mentioned in comment.

devtools/server/docs/actor-hierarchy.md
jryans: Fixed.

devtools/client/framework/target-from-url.js
jryans: Changed to one line. For whitespace generally, this is the style used in DevTools code, which is what is being modified here, so I'll leave it as-is. For code in `/positron`, I am happy to fit whatever style you prefer.

devtools/client/framework/toolbox-init.js
jryans: Good idea, added comment.

devtools/client/framework/target-from-url.js
jryans: I agree some more improvements can be made. I am going to keep the changes contained to this `window` branch for now for easier merging with m-c. I reverted the `parseInt` change outside the window path.

devtools/client/framework/toolbox-init.js
jryans: The main uses of `const` in DevTools JS are imports and actual constants, so I think I'll leave this as-is.

devtools/server/actors/root.js
jryans: It means it will destructure the name from the import. It's a common pattern in DevTools, so I don't think a comment is needed at each use.

devtools/server/actors/root.js
jryans: Fixed.

jryans:
Review comments addressed. r? @mykmelez, @ochameau.

mykmelez:
Looks great! Like @marcoscaceres, I'm a fan of *const*, but I agree that it makes sense to use it similarly to existing usage in existing files in devtools/. It looks like you've addressed all the other issues, so I'll merge this, and we can address anything else that pops up in followups.

jryans:
I filed https://bugzilla.mozilla.org/show_bug.cgi?id=1268134 to merge the DevTools changes here back to m-c, as they should be useful outside of Positron as well.

PR #17, opened 2016-04-18
define the 'global' object

mykmelez:
Define the `global` object, which is available to Node modules both via the `global` global and also implicitly, as the prototype of the module-specific global object.

There's scant documentation on this object in the Node docs themselves:

https://nodejs.org/api/globals.html#globals_global

And descriptions elsewhere are out-of-date:

http://www.hacksparrow.com/global-variables-in-node-js.html

But this implementation emulates the behavior that I see in a local copy of Node, and it appears to satisfy Electron's dependency on the object. At least it satisfies Electron's assumption that the object exists. It doesn't yet define all the properties that Electron expects to be defined on the object.

This depends on #2, so it contains all the changes on that branch. To see just the changes that were made on this branch, view the comparison between this branch and that one:

https://github.com/mykmelez/positron/compare/renderer-process-modules...global-object

### DIFF
--- a/positron/components/ModuleLoader.jsm
+++ b/positron/components/ModuleLoader.jsm
@@ -62,6 +62,35 @@ function ModuleLoader(processType) {
   const modules = new Map();
 
   /**
+   * The module-wide global object, which is available via a reference
+   * to the `global` symbol as well as implicitly via the sandbox prototype.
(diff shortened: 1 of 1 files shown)


mykmelez:
> This depends on #2, so it contains all the changes on that branch.

Rebased now that #2 has merged.

PR #24, opened 2016-04-20
fix exposure of process global to window contexts

mykmelez:
Recent changes broke exposure of the *process* global to BrowserWindow contexts. This fixes the problem.


marcoscaceres:
R+

PR #26, opened 2016-04-21
run renderer process init.js script to initialize environment (fixes #25)

### LINKED ISSUE #25: run renderer process init.js script to initialize environment
This is the renderer process half of #19.

mykmelez:
This branch evaluates the renderer process init.js script when a module loader is constructed for a new BrowserWindow. Running that script entrains a bunch of dependencies, including many Node core modules and some Node core native bindings. But relatively few of the Node APIs are actually exercised, from what I can tell. So it should be possible to get the script working without having to rewrite much Node.

So far this branch imports a bunch of Node modules verbatim and stubs a few native bindings.


positron/node/lib/_stream_duplex.js
```diff
@@ -0,0 +1,57 @@
...
+
+util.inherits(Duplex, Readable);
+
+var keys = Object.keys(Writable.prototype);
+for (var v = 0; v < keys.length; v++) {
+  var method = keys[v];
+  if (!Duplex.prototype[method])
+    Duplex.prototype[method] = Writable.prototype[method];
+}
+
```
marcoscaceres: Maybe:
```JS
Object
  .keys(Writable.prototype)
  .filter(method => !Duplex.prototype[method])
  .forEach(method => Duplex.prototype[method] = Writable.prototype[method]);
```

positron/node/lib/_stream_duplex.js
```diff
@@ -0,0 +1,57 @@
...
+function Duplex(options) {
+  if (!(this instanceof Duplex))
```
marcoscaceres: nit: please add "{" "}" to these and the ones below.

positron/node/lib/_stream_duplex.js
```diff
@@ -0,0 +1,57 @@
...
+    return new Duplex(options);
```
marcoscaceres: Is it a requirement that we allow constructors to be called without `new` here?

positron/node/lib/_stream_duplex.js
```diff
@@ -0,0 +1,57 @@
+// a duplex stream is just a stream that is both readable and writable.
+// Since JS doesn't have multiple prototypal inheritance, this class
+// prototypally inherits from Readable, and then parasitically from
```
marcoscaceres: s/parasitically from/mixes-in/

positron/node/lib/_stream_duplex.js
```diff
@@ -0,0 +1,57 @@
...
+util.inherits(Duplex, Readable);
+
+var keys = Object.keys(Writable.prototype);
+for (var v = 0; v < keys.length; v++) {
+  var method = keys[v];
+  if (!Duplex.prototype[method])
+    Duplex.prototype[method] = Writable.prototype[method];
+}
+
+function Duplex(options) {
```
marcoscaceres: I wonder if we can do:
```JS
const defaultDuplexOptions = Object.freeze({
   readable: false,
   writable: false,
   allowHalfOpen: false, 
});
function Duplex( options = defaultOptions ){
   const combinedOpts = Object.assign({}, defaultDuplexOptions, options);
   this.readable = Boolean(combinedOpts.readable);
   this.writable = Boolean(combinedOpts.readable);
   this.allowHalfOpen = Boolean(combinedOpts.allowHalfOpen);
   this.once('end', onend);
}
```

marcoscaceres:
Hmm... maybe ignore my comments as we are importing the Node modules verbatim.

mykmelez:
> Hmm... maybe ignore my comments as we are importing the Node modules verbatim.

Yes, we should avoid changing imported Node and Electron code to the extent possible. There'll be a couple cases where we do (I'm looking at you, web-view.js), but in general it's better to leave them be in order to make it easier to track upstream changes. (If you find a bug, however, then it makes sense to fix it in our repo and then submit a PR to fix it upstream as well.)

mykmelez:
Ok, I've implemented/stubbed enough to get the renderer process's init.js script to successfully evaluate up to the point that code in web-view.js throws an exception, which means we can start working on <webview>. So this is now ready for review.

@marcoscaceres are you still interested/available to review this?

Note that the code in positron/node/ is imported from the Node project, and it's unchanged, so it isn't necessary to review that code.

marcoscaceres:
Assigned to self for review. Happy to try to help.

positron/components/ModuleLoader.jsm
```diff
@@ -48,7 +45,7 @@ const windowLoaders = new WeakMap();
  * @return {Function} a module importer.
```
marcoscaceres: JSDoc comments might need updating here.

positron/components/ModuleLoader.jsm
```diff
@@ -131,20 +144,22 @@ function ModuleLoader(processType) {
...
     if (!file.exists()) {
-      throw new Error('No such module: ' + path);
+      throw new Error(`No such module: ${path} (required by ${requirer.id})`);
     }
 
     // dump('require: module found at ' + uri.spec + '\n');
 
     // Exports provided by the module.
-    let exports = Object.create(null);
+    let exports = Object.create({});
```
marcoscaceres: What's the reason for using Object.create({}) here (instead of a normal object)?

positron/components/ModuleLoader.jsm
```diff
@@ -182,18 +197,41 @@ function ModuleLoader(processType) {
...
+  //
+  // We only do this for renderer processes for now.  For browser processes,
+  // we instead require the one browser process module that renderer/init.js
+  // depends on having already been required, browser/rpc-server.js.
+  //
+  // Eventually, we should get browser/init.js working and require it here,
+  // at which point it'll require browser/rpc-server.js for us.
+  //
+  if (processType === 'renderer') {
+    this.require({}, `resource:///modules/${processType}/init.js`);
```
marcoscaceres: Nit: this one can only ever be "renderer", so it's unnecessary to make this a template string. I'd suggest just making it a normal string.

positron/modules/atom_browser_web_contents.js
```diff
@@ -33,6 +45,10 @@ let WebContents_prototype = {
     this._browserWindow._domWindow.location = url;
   },
 
+  getTitle: function() {
+    return "/* stub */";
```
marcoscaceres: maybe just return the empty string or null, and console.warn("not implemented")?

positron/electron/lib/browser/rpc-server.js
```diff
@@ -45,6 +45,13 @@ let getObjectPrototype = function(object) {
   let proto = Object.getPrototypeOf(object);
   if (proto === null || proto === Object.prototype)
     return null;
+
+  // This hack employs duck typing to compare `proto` to `Object.prototype`
+  // to work around cases in which the objects are from different sandboxes.
+  // XXX Implement a real fix for that issue and remove this hack.
+  if (Object.keys(proto).sort().join('') === Object.keys(Object.prototype).sort().join(''))
```
marcoscaceres: Is this like a ES realms thing?

positron/modules/atom_browser_web_contents.js
```diff
@@ -90,16 +106,17 @@ let WebContents_prototype = {
...
   this._browserWindow = options.browserWindow;
+  this._id = ++lastWebContentsID;
 
-  // XXX Iterate WebContents_prototype and automagically assign each property
-  // to the instance (after which we'll need to assign _getURL specially, since
-  // it's an alias for getURL).
-  this.getURL = this._getURL = WebContents_prototype.getURL;
-  this.loadURL = WebContents_prototype.loadURL;
-  this.openDevTools = WebContents_prototype.openDevTools;
+  for (let prop in WebContents_prototype) {
```
marcoscaceres: I wonder if you can `Object.assign(this, WebContents_prototype)` instead?

positron/modules/atom_common_v8_util.js
```diff
@@ -2,28 +2,11 @@
  * License, v. 2.0. If a copy of the MPL was not distributed with this
  * file, You can obtain one at http://mozilla.org/MPL/2.0/. */
 
-"use strict";
+'use strict';
```
marcoscaceres: Nit: other files use double quotes?

positron/modules/atom_renderer_ipc.js
```diff
@@ -0,0 +1,24 @@
...
+"use strict";
+
+const { classes: Cc, interfaces: Ci, results: Cr, utils: Cu } = Components;
+
+const cpmm = Cc["@mozilla.org/childprocessmessagemanager;1"].
+             getService(Ci.nsISyncMessageSender);
+
+exports.send = function(name, args) {
+  // XXX Figure out what the caller in ipc-renderer.js expects us to return.
+  return cpmm.sendAsyncMessage(name, args, { window: window });
```
marcoscaceres: nit: you can just write `{ window }` instead of `{ window: window}`

positron/modules/buffer.js
```diff
@@ -0,0 +1,12 @@
+/* This Source Code Form is subject to the terms of the Mozilla Public
+ * License, v. 2.0. If a copy of the MPL was not distributed with this
+ * file, You can obtain one at http://mozilla.org/MPL/2.0/. */
+
+'use strict';
+
+const { classes: Cc, interfaces: Ci, results: Cr, utils: Cu } = Components;
```
marcoscaceres: Nit: I'm worried about these being imported everywhere they are not used (it adds unnecessary cognitive overhead). Also, if we start linting this code, the linter will get upset that things are being defined by never used.

positron/modules/atom_common_id_weak_map.js
```diff
@@ -0,0 +1,24 @@
...
+  const map = new Map();
+  this.set = function(key, value) {
+    map.set(key, Cu.getWeakReference(value));
+  };
+  this.get = function(key) {
+    return map.get(key).get();
+  }
+  this.has = map.has.bind(map);
+  this.clear = map.clear.bind(map);
+}
```
marcoscaceres: Do you know if it matters that these properties are on the instance rather than the prototype? If we want them on the prototype, we might need something like this:

```JS 
const instances = new WeakMap();

function IDWeakMap() {
  instances.set(this, new Map());
}

IDWeakMap.prototype = {
  set(key, value) {
    instances.get(this).set(key, Cu.getWeakReference(value));
    return this;
  },
  get(key) {
    return instances.get(this).get(key).get();
  },
  has(value) {
    return instances.get(this).has(value);
  },
  clear() {
    instances.get(this).clear();
  },
};

```

positron/modules/process.js
```diff
@@ -6,8 +6,24 @@
 
 const { classes: Cc, interfaces: Ci, results: Cr, utils: Cu } = Components;
 
+// The Node version of this module only re-exports the `process` global,
+// because the global is implemented in natively.  But we implement the global
```
marcoscaceres: Typo? "implemented in natively"?

marcoscaceres:
OK, just a couple of little comments, but r+ from me otherwise.

positron/components/ModuleLoader.jsm
mykmelez: > JSDoc comments might need updating here.

Indeed, updated in 3f8c241.

positron/components/ModuleLoader.jsm
mykmelez: > What's the reason for using Object.create({}) here (instead of a normal object)?

I'm unsure why I used `Object.create(null)` instead of an object literal when I originally added this line, but the reason I changed it here to make its prototype be an object literal is that https://github.com/mozilla/positron/blob/master/positron/electron/lib/renderer/api/remote.js#L286-L289 calls `__defineGetter__` on the object, and that function isn't defined on it if its prototype is `null` instead of `Object.prototype`.

But there are simpler ways to fix that problem, like setting *exports* to an object literal. In fact it looks like we don't use it outside the *module* object, so it doesn't even need a local variable. I set module.exports directly to an object literal in e12a8c5.

positron/components/ModuleLoader.jsm
mykmelez: > Nit: this one can only ever be "renderer", so it's unnecessary to make this a template string. I'd suggest just making it a normal string.

That's true right now, but eventually I hope to be able to use the same line to load both *browser* and *renderer* init.js scripts. In the meantime, I've hardcoded the values in 7d8976f.

positron/modules/atom_browser_web_contents.js
mykmelez: > maybe just return the empty string or null, and console.warn("not implemented")?

I return the empty string in 0853030, but I didn't add a `console.warn` call, as I'd prefer to implement warnings for all stubs in #32.

positron/electron/lib/browser/rpc-server.js
mykmelez: > Is this like a ES realms thing?

I'm actually not sure what it is. I'm going to file an issue and follow up with folks. Presumably it's something to do with each sandbox having its own compartment and its own instance of globals like Object.

positron/modules/atom_common_v8_util.js
mykmelez: > Nit: other files use double quotes?

There's a mix of styles in this codebase. On this branch, the 20 JS files in positron/modules/ are evenly split between `"use strict";` and `'use strict';`. In general, I'm preferring single quotes these days, and converting files to use them, because it's the emerging standard in the Node community (and these files are Node modules, and Positron will eventually be published in NPM as a Node package).

See for example these recommendations from the top three Google search results for "node style guide":

* https://github.com/felixge/node-style-guide#use-single-quotes
* https://github.com/airbnb/javascript#strings--quotes
* https://github.com/RisingStack/node-style-guide#strings

That last link now suggests another style guide that recommends the same thing: http://standardjs.com/#rules

positron/modules/atom_browser_web_contents.js
mykmelez: > I wonder if you can Object.assign(this, WebContents_prototype) instead?

Indeed I can! Done in 6eeb841.

positron/modules/atom_renderer_ipc.js
mykmelez: > nit: you can just write { window } instead of { window: window}

Right, fixed in cfbbfd3.

positron/modules/buffer.js
mykmelez: > Nit: I'm worried about these being imported everywhere they are not used (it adds unnecessary cognitive overhead). Also, if we start linting this code, the linter will get upset that things are being defined by never used.

I agree that it's verbose, but I would fix that by injecting those globals into all native binding modules in lieu of simply providing Components, since almost all of those modules access some of those globals (or will, once implemented, for the ones that are currently stubs). I've filed #44 on that.

positron/modules/atom_common_id_weak_map.js
mykmelez: > Do you know if it matters that these properties are on the instance rather than the prototype? If we want them on the prototype, we might need something like this:

I don't think it matters. In theory, the prototype-based implementation is more performant, but I don't know if it's actually more performant in practice. And I don't think this is performance-sensitive code (i.e. that the number of IDWeakMap instances is high enough for a difference in construction perf to matter). However, I'm happy switch to it anyway, if you prefer it for another reason.

positron/modules/process.js
mykmelez: > Typo? "implemented in natively"?

Nice catch! I was wordsmithing that comment at the last minute, so I let that one slip through. Fixed in d16a2a0.

positron/electron/lib/browser/rpc-server.js
mykmelez: > I'm actually not sure what it is. I'm going to file an issue and follow up with folks. Presumably it's something to do with each sandbox having its own compartment and its own instance of globals like Object.

I filed #45 on this issue.

mykmelez:
@marcoscaceres Merge if you're satisfied with this, or let me know what else to do with it!

positron/modules/atom_common_id_weak_map.js
marcoscaceres: > On 20 May 2016, at 2:51 AM, Myk Melez <notifications@github.com> wrote:
> 
> In positron/modules/atom_common_id_weak_map.js:
> 
> > +const { classes: Cc, interfaces: Ci, results: Cr, utils: Cu } = Components;
> > +
> > +// Described In Electron's id_weak_map.h as:
> > +// "Like ES6's WeakMap, but the key is Integer and the value is Weak Pointer."
> > +
> > +function IDWeakMap() {
> > +  const map = new Map();
> > +  this.set = function(key, value) {
> > +    map.set(key, Cu.getWeakReference(value));
> > +  };
> > +  this.get = function(key) {
> > +    return map.get(key).get();
> > +  }
> > +  this.has = map.has.bind(map);
> > +  this.clear = map.clear.bind(map);
> > +}
> Do you know if it matters that these properties are on the instance rather than the prototype?
> 

It might. We should check if anyone has used the Electron version by calling the methods by going through the prototype. 

> If we want them on the prototype, we might need something like this:
> 
> I don't think it matters. In theory, the prototype-based implementation is more performant, but I don't know if it's actually more performant in practice. And I don't think this is performance-sensitive code (i.e. that the number of IDWeakMap instances is high enough for a difference in construction perf to matter). However, I'm happy switch to it anyway, if you prefer it for another reason.
> 

As above. Not really a perf concern, just a interop one. However, we could certainly see how it goes with the current implementation.

marcoscaceres:
Sent one last comment. If you are confident we can proceed wrt IDWeakMap, then let's merge.

positron/modules/atom_common_id_weak_map.js
Ryuno-Ki: FYI: Markdown isn't supported on comments by mail. I asked GitHub support in the past about it.

mykmelez: > It might. We should check if anyone has used the Electron version by calling the methods by going through the prototype.

It's intended to be an internal API, so I wouldn't expect much usage outside Electron, which doesn't itself call these methods via the prototype. This GitHub search finds some Electron forks (including ours), plus a couple of Node forks (that presumably use a different implementation—ours is from Electron, not Node), but none of those use the prototype either:

https://github.com/search?q=IDWeakMap&type=Code&utf8=%E2%9C%93

Nevertheless, it's trivial to change to the implementation that preserves the prototype chain a caller would expect, so I'll make the change!

marcoscaceres: Thanks for looking into it!

PR #27, opened 2016-04-22
ensure browserWindow on domwindowclosed

mykmelez:
After merging @jryans's devtools branch, I noticed this error when quitting the "hello world" app: `JavaScript error: … resource:///modules/gecko/atom_browser_window.js, line 71: TypeError: browserWindow is undefined`. Presumably it's because the observer gets a notification about the devtools window closing, but that window isn't a BrowserWindow, so it isn't found in `browserWindows`. Thus this branch ensures *browserWindow* before trying to use it.

### DIFF
--- a/positron/modules/atom_browser_window.js
+++ b/positron/modules/atom_browser_window.js
@@ -68,6 +68,15 @@ windowWatcher.registerNotification(function observe(subject, topic, data) {
       break;
     case 'domwindowclosed': {
       let browserWindow = browserWindows.get(subject);
+      if (!browserWindow) {
+        // The window might be a devtools windows, in which case it won't be
+        // in our list of browser windows.
+        // TODO: figure out if window-all-closed respects devtools windows,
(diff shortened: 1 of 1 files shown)


positron/modules/atom_browser_window.js
```diff
@@ -68,6 +68,15 @@ windowWatcher.registerNotification(function observe(subject, topic, data) {
       break;
     case 'domwindowclosed': {
       let browserWindow = browserWindows.get(subject);
+      if (!browserWindow) {
+        // The window might be a devtools windows, in which case it won't be
```
jryans: Nit: "devtools window"

jryans:
This seems like a reasonable fix for the issue to me. We can always adjust it as we learn more.

positron/modules/atom_browser_window.js
mykmelez: > Nit: "devtools window"

@jryans Good catch, fixed in 1c4a5e4!

PR #28, opened 2016-04-25
merge gecko-dev

mykmelez:
This merges from upstream, and the merge itself was clean, but starting the "hello world" app throws:

> JavaScript error: resource://devtools/shared/Loader.jsm, line 195: Error: Module `devtools/client/main` is not found at resource://devtools/client/main.js

And the devtools window doesn't open. It looks like this is a regression from [bug 1201475](https://bugzilla.mozilla.org/show_bug.cgi?id=1201475), although I'm unsure how to fix it. The main module appears to have been moved (with a few changes) to devtools/client/dom/main, but changing the path to the module in the require() call isn't sufficient to resolve the regression, as that triggers:

> JavaScript error: resource://gre/modules/commonjs/toolkit/loader.js, line 299: ReferenceError: `Components` is not available in this context.

@jryans, @ochameau: Do y'all have any ideas what needs to be done here?


jryans:
Taking a look now!

jryans:
@mykmelez, I think the following would fix it, but I'm waiting on build to confirm:

```udiff
diff --git a/positron/modules/atom_browser_web_contents.js b/positron/modules/atom_browser_web_contents.js
index 9b00839..eace1e6 100644
--- a/positron/modules/atom_browser_web_contents.js
+++ b/positron/modules/atom_browser_web_contents.js
@@ -35,17 +35,17 @@ let WebContents_prototype = {

   openDevTools() {
     // TODO: When tools can be opened inside the content window, support
     // `detach` option to force into a new window instead.

     // Ensure DevTools core modules are loaded, including support for the about
     // URL below which is registered dynamically.
     const { loader } = Cu.import("resource://devtools/shared/Loader.jsm", {});
-    loader.require("devtools/client/main");
+    loader.require("devtools/client/framework/devtools-browser");

     // The current approach below avoids the need for a container window
     // wrapping a tools frame, but it does replicate close handling, etc.
     // Historically we would have used toolbox-hosts.js to handle this, but
     // DevTools will be moving away from that, and so it seems fine to
     // experiment with toolbox management here.
     let window = this._browserWindow._domWindow;
     let id = window.QueryInterface(Ci.nsIInterfaceRequestor)
```

jryans:
My build confirms that appears to fix the issue.

mykmelez:
> My build confirms that appears to fix the issue.

Thanks, @jryans! Fixed in f015b91.

PR #31, opened 2016-04-30
Close DevTools window on browser close

### LINKED ISSUE #29: "hello world" app doesn't quit when I close its window
When I close the BrowserWindow in the "hello world" app, the devtools window also closes, but the app doesn't quit. On Mac, that's expected, because the main script leaves the app running on Mac after all windows have closed; but I see this on Linux too (Ubuntu 14.04).

jryans:
This fixes #29 so that calls to `quit()` after the browser window closes correctly ends the application on Linux (and hopefully Windows as well).

There is still a bit of mystery, as I don't entirely follow why the `unload` handler is needed to make this work. Happy to revise if anyone has ideas!

r? @mykmelez

### DIFF
--- a/positron/modules/atom_browser_app.js
+++ b/positron/modules/atom_browser_app.js
@@ -11,7 +11,13 @@ Cu.import('resource://gre/modules/Services.jsm');
 exports.app = {
   quit() {
     // XXX Emit the before-quit and will-quit events.
-    Services.startup.quit(Services.startup.eAttemptQuit);
+    // Wait one turn of the event loop for ancillary windows like DevTools to
+    // close first.
+    Services.tm.mainThread.dispatch({
+      run() {
+        Services.startup.quit(Services.startup.eAttemptQuit);
+      }
+    }, Ci.nsIThread.DISPATCH_NORMAL);
   },
 };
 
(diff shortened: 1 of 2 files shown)


mykmelez:
> There is still a bit of mystery, as I don't entirely follow why the unload handler is needed to make this work. Happy to revise if anyone has ideas!

Hmm, indeed, that's strange!

I just built this on Windows, and it doesn't seem to work for me there. :-(

What if *app.quit* calls `Services.startup.quit(Services.startup.eAttemptQuit)` in a timeout to give the devtools window an extra turn of the event loop to close?

jryans:
@mykmelez, does this version work on Windows? It works for me on Linux. It's a similar idea to the timeout, but uses `dispatch` which is idiom I've seen more often in Gecko for `setTimeout(0, func)` type things.

mykmelez:
> @mykmelez, does this version work on Windows? It works for me on Linux.
It works for me on Windows!

> It's a similar idea to the timeout, but uses dispatch which is the idiom I've seen more often in Gecko for setTimeout(0, func) type things.

Makes sense, let's use the common idiom!

PR #40, opened 2016-05-16
enable remote debugging of Positron itself

mykmelez:
@jryans With this branch, I can successfully open a devtools window in a separate process by passing --jsdebugger on the command line and then debug Positron chrome by setting breakpoints and stepping through the code.

The hack in jsconsole-clhandler.manifest is pretty awful, though, so I'd love your thoughts on a real fix that we can upstream!

### DIFF
--- a/positron/components/PositronCLH.manifest
+++ b/positron/components/PositronCLH.manifest
@@ -1,3 +1,3 @@
 component {dd20d954-f2a9-11e5-aeae-782bcb9e2c3f} PositronCLH.js
 contract @mozilla.org/positron/clh;1 {dd20d954-f2a9-11e5-aeae-782bcb9e2c3f}
-category command-line-handler a-positron @mozilla.org/positron/clh;1
+category command-line-handler x-positron @mozilla.org/positron/clh;1
--- a/positron/components/PositronCLH.manifest
+++ b/positron/components/PositronCLH.manifest
@@ -1,3 +1,3 @@
 component {dd20d954-f2a9-11e5-aeae-782bcb9e2c3f} PositronCLH.js
 contract @mozilla.org/positron/clh;1 {dd20d954-f2a9-11e5-aeae-782bcb9e2c3f}
-category command-line-handler x-positron @mozilla.org/positron/clh;1
+category command-line-handler a-positron @mozilla.org/positron/clh;1
--- a/toolkit/components/console/jsconsole-clhandler.manifest
+++ b/toolkit/components/console/jsconsole-clhandler.manifest
@@ -1,3 +1,3 @@
 component {2cd0c310-e127-44d0-88fc-4435c9ab4d4b} jsconsole-clhandler.js
 contract @mozilla.org/toolkit/console-clh;1 {2cd0c310-e127-44d0-88fc-4435c9ab4d4b}
-category command-line-handler b-jsconsole @mozilla.org/toolkit/console-clh;1
+category command-line-handler t-jsconsole @mozilla.org/toolkit/console-clh;1
--- a/toolkit/components/console/jsconsole-clhandler.manifest
+++ b/toolkit/components/console/jsconsole-clhandler.manifest
@@ -1,13 +1,3 @@
-# Disable jsconsole-clhandler so it doesn't override devtools-startup.
-# Firefox depends on devtools-startup overriding jsconsole-clhandler because
-# it puts it into the components/browser/ subdirectory, which is apparently
-# registered after the top-level components directory.  But Positron puts
-# devtools-startup into the top-level directory, so jsconsole-clhandler
-# overrides it.  This hack prevents it from doing so.
-#
-# XXX Figure out a better way to prevent jsconsole-clhandler from overriding
-# devtools-startup.
-#
-# component {2cd0c310-e127-44d0-88fc-4435c9ab4d4b} jsconsole-clhandler.js
-# contract @mozilla.org/toolkit/console-clh;1 {2cd0c310-e127-44d0-88fc-4435c9ab4d4b}
+component {2cd0c310-e127-44d0-88fc-4435c9ab4d4b} jsconsole-clhandler.js
+contract @mozilla.org/toolkit/console-clh;1 {2cd0c310-e127-44d0-88fc-4435c9ab4d4b}
 category command-line-handler b-jsconsole @mozilla.org/toolkit/console-clh;1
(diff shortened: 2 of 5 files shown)


mykmelez:
Incidentally, this branch outputs an error and a warning to the terminal:

>JavaScript error: resource://gre/modules/XPCOMUtils.jsm, line 191: TypeError: can't redefine non-configurable property "Services"
>JavaScript warning: resource://gre/modules/commonjs/toolkit/loader.js -> resource://devtools/shared/builtin-modules.js, line 153: DebuggeeWouldRun: debuggee `resource://gre/modules/commonjs/toolkit/loader.js:598' would run

(The warning also appears in both devtools consoles, the one opened by the "hello world" app and the one in the separate process opened by --jsdebugger.)

Neither appears to prevent the app from running, nor the devtools window from working, for that matter.

mykmelez:
> The hack in jsconsole-clhandler.manifest is pretty awful, though, so I'd love your thoughts on a real fix that we can upstream!

One possibility is to register devtools-startup with a unique contract ID and register it in the command-line-handler category with an "entry" that ensures it'll handle the command line before jsconsole-clhandler gets the opportunity, per:

https://dxr.mozilla.org/mozilla-central/source/toolkit/components/commandlines/nsICommandLineHandler.idl#17-24

mykmelez:
@jryans: df6553e makes the Positron command-line handler wait for the devtools process to connect before starting the app, accomplishing the same goal as #41, but without modifying *devtools-startup*. This implementation also doesn't spin the event loop. But it does require the Positron CLH to load before *devtools-startup*, so it can detect the `--jsdebugger` flag before *devtools-startup* removes it from the list.

And since *devtools-startup* overrides `@mozilla.org/toolkit/console-clh;1`, whose category entry is `b-jsconsole`, I had to change Positron's entry from `x-positron` to `a-positron`, which means it runs before most or all other command-line handlers. I'm unsure if there are side-effects to that change.

I'm also unsure why implementation I like better. This one doesn't spin the event loop, and it doesn't require devtools changes. But it does rely on the brittle ordering of command-line handlers. And on the details of *devtools-startup*. Whereas #41 spins the event loop but also encapsulates the code to handle `--jsdebugger` in *devtools-startup*, where it's easier to maintain across changes to that code.

jryans:
> Incidentally, this branch outputs an error and a warning to the terminal:
> 
>     JavaScript error: resource://gre/modules/XPCOMUtils.jsm, line 191: TypeError: can't redefine non-configurable property "Services"
>     JavaScript warning: resource://gre/modules/commonjs/toolkit/loader.js -> resource://devtools/shared/builtin-modules.js, line 153: DebuggeeWouldRun: debuggee `resource://gre/modules/commonjs/toolkit/loader.js:598' would run
> 
> (The warning also appears in both devtools consoles, the one opened by the "hello world" app and the one in the separate process opened by --jsdebugger.)
> 
> Neither appears to prevent the app from running, nor the devtools window from working, for that matter.

These warnings appear in Firefox's browser toolbox as well, so should be safe to ignore.

toolkit/components/console/jsconsole-clhandler.manifest
```diff
@@ -1,3 +1,13 @@
...
+# registered after the top-level components directory.  But Positron puts
+# devtools-startup into the top-level directory, so jsconsole-clhandler
+# overrides it.  This hack prevents it from doing so.
+#
+# XXX Figure out a better way to prevent jsconsole-clhandler from overriding
+# devtools-startup.
+#
+# component {2cd0c310-e127-44d0-88fc-4435c9ab4d4b} jsconsole-clhandler.js
+# contract @mozilla.org/toolkit/console-clh;1 {2cd0c310-e127-44d0-88fc-4435c9ab4d4b}
 category command-line-handler b-jsconsole @mozilla.org/toolkit/console-clh;1
```
jryans: Let's revert this back to what it was, except change it's entry so it runs later. Looking across the current [landscape][1] of handler entries, maybe something like `t-jsconsole` (t for toolkit), which would be near the bottom, but still before things like `x-*` and `y-*` which are trying be default / final handlers.

Then let's add a new category entry for `devtools-startup.js` (with its own contract ID as well) which uses the "normal" priority "m", something like `m-devtools`.

This should give the right effect for Positron and maintain it for Firefox: `devtools-startup` runs before the toolkit handler. It does mean that `--jsconsole` appears twice in the help output, but that seems acceptable I think.

I wanted to test my theories, so I opened #42 for this part.

[1]: https://dxr.mozilla.org/mozilla-central/search?q=command-line-handler&redirect=false&case=true

positron/components/PositronCLH.manifest
```diff
@@ -1,3 +1,3 @@
 component {dd20d954-f2a9-11e5-aeae-782bcb9e2c3f} PositronCLH.js
 contract @mozilla.org/positron/clh;1 {dd20d954-f2a9-11e5-aeae-782bcb9e2c3f}
-category command-line-handler x-positron @mozilla.org/positron/clh;1
+category command-line-handler a-positron @mozilla.org/positron/clh;1
```
jryans: It seems fine to have the application CLH run early like this, as far as I can tell. Firefox actually break theirs into two, one at `m-browser` and another at `x-default`. It seems like people are just making up names as needed to get the ordering right...

mykmelez: > It seems fine to have the application CLH run early like this, as far as I can tell. Firefox actually break theirs into two, one at m-browser and another at x-default. It seems like people are just making up names as needed to get the ordering right...

Sounds good. Although now that devtools is at m-devtools, I'll make this something like e-positron, so it can be overridable by something else if needed.

jryans:
Overall, I think this looks good, but I prefer the startup debugging approach from #41 since it seems to work for Firefox as well, so let's use that route for the startup portion.

toolkit/components/console/jsconsole-clhandler.manifest
mykmelez: > Let's revert this back to what it was, except change it's entry so it runs later.
> …
> I wanted to test my theories, so I opened #42 for this part.

Sounds good, I merged #42 and reverted this change in 6128262.

mykmelez:
> Overall, I think this looks good, but I prefer the startup debugging approach from #41 since it seems to work for Firefox as well, so let's use that route for the startup portion.

Ok, sounds good, I've reverted the code on this branch to wait for the jsdebugger in 13e057e.

positron/components/PositronCLH.manifest
mykmelez: > Sounds good. Although now that devtools is at m-devtools, I'll make this something like e-positron, so it can be overridable by something else if needed.

Erm, actually I don't need to change the entry for the Positron CLH at all anymore, since we're taking the implementation in #41. So I've left it at `m-positron`.

mykmelez: Erm, I mean `x-positron`. 😴

jryans:
Great, this looks ready to merge!

PR #41, opened 2016-05-17
pause until debugger thread resumed so we can debug startup

mykmelez:
@jryans This works as expected when applied on top of #40. Is it something that DevTools might be willing take upstream? Perhaps in coordination with another flag like `--pause-jsdebugger-until-server-thread-resumes-even-if-that-means-hanging-the-app`?

Note: I do see two instances of this error after the server connects:

>Handler function threw an exception: TypeError: this._progressListener is undefined
>Stack: _onDocShellCreated/<@resource://gre/modules/commonjs/toolkit/loader.js -> resource://devtools/server/actors/webbrowser.js:1319:9
>exports.makeInfallible/<@resource://gre/modules/commonjs/toolkit/loader.js -> resource://devtools/shared/ThreadSafeDevToolsUtils.js:101:14
>Line: 1319, column: 9

### DIFF
--- a/devtools/server/actors/webbrowser.js
+++ b/devtools/server/actors/webbrowser.js
@@ -1316,7 +1316,12 @@ TabActor.prototype = {
       // In child processes, we have new root docshells,
       // let's watch them and all their child docshells.
       if (this._isRootDocShell(docShell)) {
-        this._progressListener.watch(docShell);
+        // Don't assume there's a progress listener, since Positron starts
+        // without a window, so this._progressListener never gets created.
+        // TODO: ensure there's a listener for apps without an initial window.
+        if (this._progressListener) {
+          this._progressListener.watch(docShell);
+        }
       }
       this._notifyDocShellsUpdate([docShell]);
     });
--- a/devtools/server/actors/webbrowser.js
+++ b/devtools/server/actors/webbrowser.js
@@ -1399,7 +1399,9 @@ TabActor.prototype = {
     // Stop watching this docshell (the unwatch() method will check if we
     // started watching it before).
     webProgress.QueryInterface(Ci.nsIDocShell);
-    this._progressListener.unwatch(webProgress);
+    if (this._progressListener) {
+      this._progressListener.unwatch(webProgress);
+    }
 
     if (webProgress.DOMWindow == this._originalWindow) {
       // If the original top level document we connected to is removed,
@@ -1480,7 +1482,9 @@ TabActor.prototype = {
     // Check for docShell availability, as it can be already gone
     // during Firefox shutdown.
     if (this.docShell) {
-      this._progressListener.unwatch(this.docShell);
+      if (this._progressListener) {
+        this._progressListener.unwatch(this.docShell);
+      }
       this._restoreDocumentSettings();
     }
     if (this._progressListener) {
--- a/devtools/client/devtools-startup.js
+++ b/devtools/client/devtools-startup.js
@@ -17,7 +17,7 @@
 
 "use strict";
 
-const { interfaces: Ci, utils: Cu } = Components;
+const { classes: Cc, interfaces: Ci, utils: Cu } = Components;
 const kDebuggerPrefs = [
   "devtools.debugger.remote-enabled",
   "devtools.chrome.enabled"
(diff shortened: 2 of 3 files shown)


mykmelez:
Upon further reflection, I might be able to simply delay parsing the *appPath* argument, and starting the app, until `devtools-thread-resumed`. Then I wouldn't need to spin the event loop.

mykmelez:
> Upon further reflection, I might be able to simply delay parsing the appPath argument, and starting the app, until devtools-thread-resumed. Then I wouldn't need to spin the event loop.

I tried this over in #40, and it works, but it has its own drawbacks. I'm unsure which implementation I prefer.

jryans:
This approach to startup debugging seems better to me, since it would be available to any application. I applied this to a Firefox build, and it enabled startup debugging over there as well, which is great!

I think it should be fine to upstream this mostly as-is, but @ochameau and I would need to work out how to handle the `_progressListener` error. This is happening because we currently [expect a window to exist][1] at startup if there are ever going to be windows opening at all. @ochameau and I can work that out in a bug to merge this upstream.

The `_progressListener` is there to notice if windows change location, etc. which is not too important for Positron's platform debugging. So for Positron at least, you could silence the error if you like by changing `webbrowser.js` to test for `this._progressListener` first [where the error happens][2]. That should be enough to merge this approach now for Positron without waiting on upstream.

[2]: https://dxr.mozilla.org/mozilla-central/rev/a884b96685aa13b65601feddb24e5f85ba861561/devtools/server/actors/webbrowser.js#1319

[1]: https://dxr.mozilla.org/mozilla-central/rev/a884b96685aa13b65601feddb24e5f85ba861561/devtools/server/actors/webbrowser.js#1195

mykmelez:
> The _progressListener is there to notice if windows change location, etc. which is not too important for Positron's platform debugging. So for Positron at least, you could silence the error if you like by changing webbrowser.js to test for this._progressListener first where the error happens. That should be enough to merge this approach now for Positron without waiting on upstream.

Ok, sounds good, I'll make that change!

mykmelez:
> Ok, sounds good, I'll make that change!

I made the change in 4735aeb.

devtools/server/actors/webbrowser.js
```diff
@@ -1316,7 +1316,12 @@ TabActor.prototype = {
       // In child processes, we have new root docshells,
       // let's watch them and all their child docshells.
       if (this._isRootDocShell(docShell)) {
-        this._progressListener.watch(docShell);
+        // Don't assume there's a progress listener, since Positron starts
+        // without a window, so this._progressListener never gets created.
+        // TODO: ensure there's a listener for apps without an initial window.
+        if (this._progressListener) {
```
jryans: I spotted a few more cases where you might want the same thing:

https://dxr.mozilla.org/mozilla-central/rev/f3f2fa1d7eed5a8262f6401ef18ff8117a3ce43e/devtools/server/actors/webbrowser.js#1397
https://dxr.mozilla.org/mozilla-central/rev/f3f2fa1d7eed5a8262f6401ef18ff8117a3ce43e/devtools/server/actors/webbrowser.js#1478

jryans:
Overall this looks good to me! You may want to add the extra checks I just commented for safety, but I wasn't sure whether they'd be hit in Positron or not.

mykmelez:
> Overall this looks good to me! You may want to add the extra checks I just commented for safety, but I wasn't sure whether they'd be hit in Positron or not.

I hit one of them earlier this morning!

> JavaScript error: resource://gre/modules/commonjs/toolkit/loader.js -> resource://devtools/server/actors/webbrowser.js, line 1402: TypeError: this._progressListener is undefined

(Line number slightly different, presumably because of the changes I've already made to this file, but this is the same as the first additional case you noted.)

Additional cases fixed in d6e3096.

jryans:
Great, this looks ready to go!

jryans:
I'll file a bug to upstream in a bit.

mykmelez:
> I'll file a bug to upstream in a bit.

Thanks, @jryans! Please cc: me on it!

jryans:
This will be merged as part of https://bugzilla.mozilla.org/show_bug.cgi?id=1275942.

PR #42, opened 2016-05-17
Include DevTools command line handler via category entry

jryans:
@mykmelez, here's the `command-line-handler` ordering that should be upstreamable.

It seems to work for Positron and Firefox in my testing.


mykmelez:
LGTM! And seems a lot more sensible than surreptitiously hijacking the console-clh contract.

mykmelez:
@jryans Would you like to upstream this, or shall I?

jryans:
I can upstream it, I filed https://bugzilla.mozilla.org/show_bug.cgi?id=1273967 to do so.

PR #47, opened 2016-05-20
define IDWeakMap methods on its prototype

mykmelez:
@marcoscaceres The only difference from your suggested implementation is that *set* doesn't return `this`, because the upstream implementation doesn't do that:

https://github.com/electron/electron/blob/v0.37.8/atom/common/api/atom_api_id_weak_map.cc#L21-L25


marcoscaceres:
r+ from me. It's not great that this doesn't follow the WeakMap API behaviour, but it's ok given that it's just internal.

PR #50, opened 2016-05-23
ensure iframe first; set toolboxOpened unconditionally; end after catch

mykmelez:
@jryans While resolving conflicts with a merge from upstream, I noticed a few issues with toolbox-init:

1. The script tests `if (iframe)` after calling `iframe.QueryInterface()`, but that call should throw an exception if *iframe* is falsy, so the conditional will never be false.
2. After fixing issue #1 by testing *iframe* before calling `iframe.QueryInterface()`, if *iframe* is falsy, then *toolboxOpened* is never defined, so the `toolboxOpened.catch()` call below will fail.
3. The promise chain will reach the `toolboxOpened.then()` function even if there was an error opening the toolbox, since it gets attached to the chain after `toolboxOpened.catch()`, which doesn't re-throw the error.

Here are fixes for these issues.

### DIFF
--- a/devtools/client/framework/toolbox-init.js
+++ b/devtools/client/framework/toolbox-init.js
@@ -50,12 +50,12 @@ if (url.search.length > 1) {
     // `iframe` is the targeted document to debug
     let iframe = host.wrappedJSObject ? host.wrappedJSObject.target
                                       : host.target;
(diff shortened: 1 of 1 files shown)


jryans:
Looks good to me, thanks!

PR #51, opened 2016-05-24
make the window object be the global object for modules in renderer processes

mykmelez:
@marcoscaceres This is another step toward implementing *&lt;webview>*, as the modules that implement it (and the other modules that get loaded in the renderer process) all assume that their global object is the *DOMWindow* object for the renderer window, and they access a variety of *DOMWindow* built-ins on it.

Instead of identifying each individual built-in those modules use and then manually adding them to the global that the *ModuleLoader* creates, this branch makes the *DOMWindow* object itself be the global, so all of its built-ins are available to renderer process modules.

Doing this means that the *process* global for those objects is now the one we bind to *DOMWindow* via WebIDL, and that entrained some changes to the way we expose its properties and methods Basically: we had to explicitly expose them via WebIDL, which we needed to do eventually anyway (in order for the web page itself to have access to them).

### DIFF
--- a/positron/components/ModuleLoader.jsm
+++ b/positron/components/ModuleLoader.jsm
@@ -13,7 +13,6 @@ const Ci = Components.interfaces;
 const Cr = Components.results;
 const Cu = Components.utils;
 
-Cu.import('resource://gre/modules/Console.jsm');
 Cu.import('resource://gre/modules/Services.jsm');
 
 const subScriptLoader = Cc['@mozilla.org/moz/jssubscript-loader;1'].
@@ -52,6 +51,14 @@ const windowLoaders = new WeakMap();
  */
 
 function ModuleLoader(processType, window) {
+  // Register this loader for the given window early, so modules that we require
+  // during loader construction can access window.process without causing
+  // the `process` WebIDL binding to create a new module loader (and thus enter
+  // an infinite loop).
+  if (window) {
+    windowLoaders.set(window, this);
+  }
+
   const globalPaths = DEFAULT_GLOBAL_PATHS.slice();
   // Prepend a process-type-specific path to the global paths for this loader.
   globalPaths.unshift(processType + '/api/exports/');
@@ -69,26 +76,18 @@ function ModuleLoader(processType, window) {
    * to the `global` symbol as well as implicitly via the sandbox prototype.
    * This object is shared across all modules loaded by this loader.
    */
-  this.global = {
-    // Electron assumes these exist on the global object, so we define them.
-    // I wonder if it assumes that all window.* properties will exist, in which
-    // case we should instead make the window be the global's prototype.
-    window: processType === 'renderer' ? window : undefined,
-    location: processType === 'renderer' ? window.location : undefined,
-    document: processType === 'renderer' ? window.document : undefined,
-
-    // This comes from Console.jsm, and I don't think it's exactly the same
-    // as the Console Web API.
-    // XXX Replace this with the real Console Web API (or at least ensure
-    // that the current version is compatible with it).
-    console: console,
-
-    // Other global properties, like `process` and `Buffer`, are defined
-    // after we define the require() function, so we can use that function
-    // to implement them.
-
-    // XXX Also define setImmediate and clearImmediate.
-  };
+  if (processType === 'browser') {
+    this.global = {
+      // I don't think this is exactly the same as the Console Web API.
+      // XXX Replace this with the real Console Web API (or at least ensure
+      // that the current version is compatible with it).
+      console: Cu.import('resource://gre/modules/Console.jsm', {}).console,
+    };
+  } else {
+    // For renderer processes, the global object is the window object
+    // of the renderer window, so modules have access to all its globals.
+    this.global = window;
+  }
 
   /**
    * Add default properties to the module-specific global object.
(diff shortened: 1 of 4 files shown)


positron/webidl/Process.webidl
```diff
@@ -3,14 +3,36 @@
...
+};
+
 [ChromeOnly,
  JSImplementation="@mozilla.org/positron/process;1"]
-interface processImpl {
+interface processImpl : EventEmitter {
+  [Cached, Pure] readonly attribute sequence<DOMString> argv;
+  [Cached, Pure] readonly attribute object env;
+  [Cached, Pure] readonly attribute DOMString execPath;
+  [Cached, Pure] readonly attribute long pid;
```
marcoscaceres: Maybe "unsigned long"? I don't know of any pids being signed on any OSs.

positron/webidl/Process.webidl
```diff
@@ -3,14 +3,36 @@
  * You can obtain one at http://mozilla.org/MPL/2.0/.
  */
 
+dictionary ReleaseDictionary {
+  DOMString name;
```
marcoscaceres: Note that node process also has:

 * `sourceUrl`
 * `headersUrl`

positron/components/ModuleLoader.jsm
```diff
@@ -200,21 +199,29 @@ function ModuleLoader(processType, window) {
...
   };
 
-  // The `process` global is complicated.  It's implemented as a module
-  // that uses `require` to import other modules, but it also needs to exist
-  // before those modules are evaluated, since it's a global that's supposed
-  // to be available in them.
+  // The `process` global is complicated.  It's implemented by a module,
+  // process.js, that we require.  But it's also supposed to exist before any
+  // modules are required, since it's a global that's supposed to be available
+  // in all of them.
```
marcoscaceres: it's not entirely clear what "them" means here. Maybe change to "in all modules".

positron/modules/process.js
```diff
@@ -16,8 +16,15 @@ Cu.import('resource://gre/modules/Services.jsm');
...
 const exeFile = Services.dirsvc.get("XREExeF", Ci.nsIFile);
 
-// Re-export `process` as a native module.  I dunno why this matters, since all
-// modules already have access to the global.  But Node does it, so we do too.
+// This file is a special kind of module, as the module loader requires it
+// during loader construction and uses it to define the `process` global.
+// Thus that global is not available at evaluation time, neither for this module
+// nor for any other modules it requires upon evaluation.
+
+const EventEmitter = require('events').EventEmitter;
```
marcoscaceres: nit:
```
const { EventEmitter } = require('events');
```

positron/modules/process.js
```diff
@@ -16,8 +16,15 @@ Cu.import('resource://gre/modules/Services.jsm');
...
+const process = new EventEmitter();
```
marcoscaceres: Q: shouldn't process be an object that has EventEmitter as its prototype?

```
const process = Object.create(EventEmitter); 
``

marcoscaceres: If the above is correct, we should really refactor the process object to define its properties in adherence to Object.create().

marcoscaceres:
Couple of questions would like to see clarified before merge. Excited to see the foundations coming together 👍

positron/webidl/Process.webidl
mykmelez: > Maybe "unsigned long"? I don't know of any pids being signed on any OSs.

Good point, fixed in 24bad52.

positron/webidl/Process.webidl
mykmelez: > Note that node process also has:

Indeed, and libUrl on Windows. This isn't a complete specification of the Node API for the *process* global, nor is the underlying implementation complete. It's just the minimum necessary to get Electron to work for the Hello World test app (and soon for a Basic Browser test app).

a7146fb adds commentary to that effect.

positron/components/ModuleLoader.jsm
mykmelez: > it's not entirely clear what "them" means here. Maybe change to "in all modules".

Yes, that's what I intended to express. I've clarified using your suggested wording in 4369062.

positron/modules/process.js
mykmelez: > const { EventEmitter } = require('events');

Yup, fixed in 92d393b.

positron/modules/process.js
mykmelez: > Q: shouldn't process be an object that has EventEmitter as its prototype?

The version of Node I'm integrating (5.x) [recommends either](https://nodejs.org/docs/latest-v5.x/api/events.html) the "traditional Node.js style prototypical inheritance using the util.inherits() method" or ES6 classes and *extends* to create a class that extends *EventEmitter*.

But the version of Electron I'm integrating uses either `var object = new EventEmitter` or `object.__proto__ = EventEmitter.prototype` to either create *EventEmitter* instances or decorate instances with *EventEmitter* properties and methods, depending on whether or not the object in question exists yet.

In the latter case, and in the case of `Object.create(EventEmitter.prototype)`, the object doesn't get initialized by the *EventEmitter* constructor, so its *_events* and *_eventsCount* properties are `undefined`. (Other initialization also doesn't take place, like for the *domain* property, although I'm unfamiliar with that property and don't know if this matters for Electron.)

Ironically, Node appears to anticipate *EventEmitter* being used that way, since its methods always check if *_events* is defined before using that property and *_eventsCount*. Which is why the current implementation, which uses the `object.__proto__ = EventEmitter.prototype` syntax to make *process* an *EventEmitter*, works as expected.

But in this branch I confirmed that I can `require('events')` at the top of process.js, so I figured it would make more sense to instantiate *process* as an *EventEmitter* to begin with, instead of assigning its *\_\_proto\_\_* to the *EventEmitter* prototype, to avoid the perf penalty of modifying the prototype of an object after instantiation. Hence `const process = new EventEmitter()`.

mykmelez: > If the above is correct, we should really refactor the process object to define its properties in adherence to Object.create().

I'm unsure what this means, but the intention here is to create a *process* singleton that extends the *EventEmitter* base functionality with the *process* object's additional properties and methods.

I could create a *Process* class or prototype and use classical or prototypal inheritance to establish the relationship between *EventEmitter* and that class/prototype, and that would make sense if I was constructing multiple instances of the *process* object, but it seems like overkill when *process* is a singleton.

marcoscaceres:
Thanks for the clarifications.

PR #52, opened 2016-05-24
make chrome devtools window resizable

mykmelez:
@jryans This makes my debugging experience moar better.

### DIFF
--- a/positron/components/PositronCLH.js
+++ b/positron/components/PositronCLH.js
@@ -97,7 +97,7 @@ PositronCLH.prototype = {
         };
         if (isLocal(resolvedURI)) {
           // If the URI is local, we are sure it won't wrongly inherit chrome privs
-          var features = "chrome,dialog=no,all";
+          var features = "chrome,dialog=no,all,resizable";
           Services.ww.openWindow(null, resolvedURI.spec, "_blank", features, null);
           cmdLine.preventDefault = true;
           return;


positron/components/PositronCLH.js
```diff
@@ -97,7 +97,7 @@ PositronCLH.prototype = {
         };
         if (isLocal(resolvedURI)) {
           // If the URI is local, we are sure it won't wrongly inherit chrome privs
-          var features = "chrome,dialog=no,all";
+          var features = "chrome,dialog=no,all,resizable";
```
jryans: Hmm, the Firefox Browser Toolbox equivalent is resizable, so I wonder why yours is not... I guess `all` implies resizable with Firefox. I'll try to see if I can find out why it differs here...

jryans: Well, that was quite a rabbit hole... So, for the "all" feature to apply correctly, `openWindow` must consider you to be a [`dialog`][1], where that is apparently defined as "passed a non-zero length args array".

And sure enough, Firefox's `nsBrowserContentHandler.js` has an [`openWindow`][2] wrapper which seems to create an args array and add at least one element down each of its various paths.

So, if you want to more closely match Firefox here, you could do something like:

```js
          var args = Components.classes["@mozilla.org/supports-array;1"]
                               .createInstance(Components.interfaces.nsISupportsArray);
          args.AppendElement(null);
          Services.ww.openWindow(null, resolvedURI.spec, "_blank", features, args);
```

Perhaps it's best to match Firefox since this should enable many [features][3], instead of listing them one by one.

I opened #53 to discuss this approach.

[2]: https://dxr.mozilla.org/mozilla-central/rev/46fe2115d46a5bb40523b8466341d8f9a26e1bdf/browser/components/nsBrowserContentHandler.js#178
[1]: https://dxr.mozilla.org/mozilla-central/rev/46fe2115d46a5bb40523b8466341d8f9a26e1bdf/embedding/components/windowwatcher/nsWindowWatcher.cpp#367
[3]: https://dxr.mozilla.org/mozilla-central/source/embedding/browser/nsIWebBrowserChrome.idl#45-55

jryans:
Looks like we can close this since #53 was merged.

mykmelez:
Closed in favor of #53.

PR #54, opened 2016-05-24
Resolve window errors when platform debugging

### LINKED ISSUE #48: Various errors in Positron debugger
We're now able to open a debugger for the Positron platform by starting with `--jsdebugger`. This opens a toolbox similar in concept to Firefox's Browser Toolbox that runs from a separate process.

However, the actors give various errors at the moment:

Console says: `resource://devtools/server/actors/webconsole.js, line 453: TypeError: argument is not a global object`
Inspector says: `TypeError: window is null`

In general, I think this is related to the shortcuts we took to allow the tools to start very early before a window appeared.

jryans:
I was able to get around the issues from #48 by ensuring we always have a window, and during early startup, we can use the hidden window for that purpose. This allows us to revert the various `_progressListener` guards that were added recently.

Overall the tools are working much better now in the platform debugger. However, I did notice that typing in the console triggers autocompletion that seems to send things over Positron's IPC mechanism, but it fails for some reason. I did not work that part out yet, but added a small debugging aid to improve the error at least. Console evaluation should now be working, so you ignore these autocomplete errors for now.

Fixes #48. @mykmelez, want to take a look?

### DIFF
--- a/positron/modules/atom_renderer_ipc.js
+++ b/positron/modules/atom_renderer_ipc.js
@@ -25,7 +25,7 @@ exports.sendSync = function(name, args) {
   // meta protocol used by Electron's RPC server.  This increases the chances we
   // log an error specific to problem site, instead of here at this IPC code.
   if (!result) {
-    result = JSON.stringify({
+    return JSON.stringify({
       type: 'value',
       value: result,
     });
--- a/positron/modules/atom_renderer_ipc.js
+++ b/positron/modules/atom_renderer_ipc.js
@@ -20,5 +20,15 @@ exports.sendSync = function(name, args) {
   // <http://electron.atom.io/docs/api/ipc-renderer/#ipcrenderersendsyncchannel-arg1-arg2->
   // expects to receive a single return value.  So this function returns
   // the first item in the array.
-  return cpmm.sendSyncMessage(name, args, { window })[0];
+  let result = cpmm.sendSyncMessage(name, args, { window })[0];
+  // If no one was listening for the message, wrap up the falsy result in the
+  // meta protocol used by Electron's RPC server.  This increases the chances we
+  // log an error specific to problem site, instead of here at this IPC code.
+  if (!result) {
+    result = JSON.stringify({
+      type: 'value',
+      value: result,
+    });
+  }
+  return result;
 };
--- a/devtools/server/actors/chrome.js
+++ b/devtools/server/actors/chrome.js
@@ -50,6 +50,13 @@ function ChromeActor(aConnection) {
   if (!window) {
     window = Services.wm.getMostRecentWindow(null);
   }
+
+  // We really want _some_ window at least, so fallback to the hidden window if
+  // there's nothing else (such as during early startup).
+  if (!window) {
+    window = Services.appShell.hiddenDOMWindow;
+  }
+
   // On xpcshell, there is no window/docshell
   let docShell = window ? window.QueryInterface(Ci.nsIInterfaceRequestor)
                                 .getInterface(Ci.nsIDocShell)
--- a/devtools/server/actors/chrome.js
+++ b/devtools/server/actors/chrome.js
@@ -130,9 +130,7 @@ ChromeActor.prototype._attach = function () {
     if (docShell == this.docShell) {
       continue;
     }
-    if (this._progressListener) {
-      this._progressListener.watch(docShell);
-    }
+    this._progressListener.watch(docShell);
   }
 };
 
@@ -155,9 +153,7 @@ ChromeActor.prototype._detach = function () {
     if (docShell == this.docShell) {
       continue;
     }
-    if (this._progressListener) {
-      this._progressListener.unwatch(docShell);
-    }
+    this._progressListener.unwatch(docShell);
   }
 
   TabActor.prototype._detach.call(this);
--- a/devtools/server/actors/webbrowser.js
+++ b/devtools/server/actors/webbrowser.js
@@ -1316,12 +1316,7 @@ TabActor.prototype = {
       // In child processes, we have new root docshells,
       // let's watch them and all their child docshells.
       if (this._isRootDocShell(docShell)) {
-        // Don't assume there's a progress listener, since Positron starts
-        // without a window, so this._progressListener never gets created.
-        // TODO: ensure there's a listener for apps without an initial window.
-        if (this._progressListener) {
-          this._progressListener.watch(docShell);
-        }
+        this._progressListener.watch(docShell);
       }
       this._notifyDocShellsUpdate([docShell]);
     });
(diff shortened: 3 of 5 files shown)


mykmelez:
> Overall the tools are working much better now in the platform debugger. However, I did notice that typing in the console triggers autocompletion that seems to send things over Positron's IPC mechanism, but it fails for some reason. I did not work that part out yet, but added a small debugging aid to improve the error at least. Console evaluation should now be working, so you ignore these autocomplete errors for now.

The reason for this is quite complex. For renderer processes, I implemented Node's *process* global via a [WebIDL binding](https://github.com/mozilla/positron/blob/master/positron/webidl/Process.webidl), because it needs to be available not only to modules but also the page loaded into the window.

That means it gets added to every *DOMWindow*, including the hidden window! However, it's lazily instantiated, so it doesn't get instantiated when the hidden window loads. Once you evaluate something in the devtools console, which is hooked up to the hidden window, however, the console does something (iterate properties?) that causes *process* to be instantiated, at which point:

1. the WebIDL binding creates an instance of the [Process XPCOM component](https://github.com/mozilla/positron/blob/master/positron/components/Process.js);
2. the Process instance calls [ModuleLoader.getLoaderForWindow](https://github.com/mozilla/positron/blob/master/positron/components/ModuleLoader.jsm#L234-L241), which creates an instance of *ModuleLoader* for the hidden window;
3. the *ModuleLoader* instance loads [Electron's renderer/init.js](https://github.com/mozilla/positron/blob/master/positron/electron/lib/renderer/init.js) to initialize the Electron environment;
4. initialization includes [this call in override.js](https://github.com/mozilla/positron/blob/master/positron/electron/lib/renderer/override.js#L11) to `remote.getCurrentWindow()`;
5. [getCurrentWindow in remote.js](https://github.com/mozilla/positron/blob/master/positron/electron/lib/renderer/api/remote.js#L271-L274) calls `ipcRenderer.sendSync('ATOM_BROWSER_CURRENT_WINDOW')`;
6. that triggers IPC to the main process, where it's [handled by BrowserWindow.prototype.receiveMessage](https://github.com/mozilla/positron/blob/master/positron/modules/atom_browser_window.js#L80-L113), which returns undefined because the window from which the message was sent is not the window to which the particular instance of *BrowserWindow* belongs.

So it never reaches [the ATOM_BROWSER_CURRENT_WINDOW handler in rpc-server.js](https://github.com/mozilla/positron/blob/master/positron/electron/lib/browser/rpc-server.js#L273-L279), although I don't think that handler would successfully return a window if the message did make it that far, since there isn't a BrowserWindow (nor WebContents) for the hidden window.

There are a couple lessons here.

First, since *ipcRenderer.sendSync* only expects to receive a single return value, we should ensure that *BrowserWindow.prototype.receiveMessage* also returns a single value from the correct *BrowserWindow* instance. Currently, it'll return an array of values, one for each *BrowserWindow* instance, and atom_renderer_ipc always returns the first item in the array, even though the actual return value may be a subsequent item. (We get away with that for now because we've only been testing apps with one BrowserWindow.) Perhaps *BrowserWindow.prototype.receiveMessage* should be a static method that receives all IPC messages and routes them to the correct BrowserWindow instance.

Second, we should ensure that the hidden window doesn't try to initialize an Electron environment, since that environment should only exist in the windows that the Electron app opens. It might be possible to restrict the *process* global via the WebIDL binding with a [Func attribute](https://developer.mozilla.org/en-US/docs/Mozilla/WebIDL_bindings#Func), although it looks like we'd have to [implement it in C++](https://developer.mozilla.org/en-US/docs/Mozilla/WebIDL_bindings#Checking_for_Permissions_or_Preferences), even though the *Process* component is in JS. Alternately, the *Process* component might be able to work around the problem by not created a *ModuleLoader* instance. Then *process* would still exist in the hidden window, but it wouldn't trigger Electron initialization.

mykmelez:
> First, since ipcRenderer.sendSync only expects to receive a single return value, we should ensure that BrowserWindow.prototype.receiveMessage also returns a single value from the correct BrowserWindow instance. Currently, it'll return an array of values, one for each BrowserWindow instance, and atom_renderer_ipc always returns the first item in the array, even though the actual return value may be a subsequent item. (We get away with that for now because we've only been testing apps with one BrowserWindow.) Perhaps BrowserWindow.prototype.receiveMessage should be a static method that receives all IPC messages and routes them to the correct BrowserWindow instance.

I've filed issue #55 on this problem, since it's unrelated to the problem you're solving here, and your changes merely expose this issue.

> Second, we should ensure that the hidden window doesn't try to initialize an Electron environment, since that environment should only exist in the windows that the Electron app opens. It might be possible to restrict the process global via the WebIDL binding with a Func attribute, although it looks like we'd have to implement it in C++, even though the Process component is in JS. Alternately, the Process component might be able to work around the problem by not created a ModuleLoader instance. Then process would still exist in the hidden window, but it wouldn't trigger Electron initialization.

This one isn't entirely related either, but I'd like to at least work around it in this changeset, even if we don't completely fix it.

positron/modules/atom_renderer_ipc.js
```diff
@@ -20,5 +20,15 @@ exports.sendSync = function(name, args) {
   // <http://electron.atom.io/docs/api/ipc-renderer/#ipcrenderersendsyncchannel-arg1-arg2->
   // expects to receive a single return value.  So this function returns
   // the first item in the array.
-  return cpmm.sendSyncMessage(name, args, { window })[0];
+  let result = cpmm.sendSyncMessage(name, args, { window })[0];
+  // If no one was listening for the message, wrap up the falsy result in the
+  // meta protocol used by Electron's RPC server.  This increases the chances we
+  // log an error specific to problem site, instead of here at this IPC code.
+  if (!result) {
+    result = JSON.stringify({
```
marcoscaceres: is there some reason to not create a new variable here? It's a bit confusing how `value` is set to `result`, but then result gets trashed by the result of JSON.stringify. Will make things hard to debug in the future if need be.

mykmelez:
> This one isn't entirely related either, but I'd like to at least work around it in this changeset, even if we don't completely fix it.

I implemented a prospective workaround in https://github.com/jryans/positron/pull/3.

positron/modules/atom_renderer_ipc.js
jryans: No particular reason, I can use `return` here instead. Fixed in 782b63834d10e4769baedbf7ca8e56db45ac9d16.

jryans:
@mykmelez, I've rolled in your fix and addressed the other review comments. What do you think?

mykmelez:
> @mykmelez, I've rolled in your fix and addressed the other review comments. What do you think?

Looks great!

jryans:
This will be upstreamed as part of https://bugzilla.mozilla.org/show_bug.cgi?id=1275942.

PR #57, opened 2016-05-26
warn once for each stub (fixes #32)

### LINKED ISSUE #32: make module and function stubs warn once per session upon evaluation/invocation
We've stubbed a variety of modules and the functions they expose, but none of the stubs warn that they're stubbly, which is dangerous, because it makes them easy to overlook when troubleshooting an issue that may be caused by an incomplete or missing implementation.

So we should make those stubs warn (once per session) when the stubbed modules are evaluated and the stubbed functions are invoked.

mykmelez:
This fixes #32 by adding _positronUtil.makeStub()_, which makes a "stub function" that warns the first time it's called, and then replacing all `function() { /* stub */ }` occurrences with functions created by *makeStub()*.

Modules that are themselves stubs call *makeStub()()* to make a stub function and then immediately call it to output a warning (so we see the warning when the module is first evaluated).

Stub functions that need to return a value pass the return value to *makeStub()*, which then returns it each time the stub is called (although it still only warns once). Stubs that need to generate a return value (or do something else when the stub is called) pass a function argument for the *returnValue* property, which the "warn once" function will then call to generate the return value.

This is essentially the same as [addUnimplementedNative() from PluotSorbet](https://github.com/mozilla/pluotsorbet/blob/16cc4934a4a1dcfeeb2ccc5102aa5cfec768370d/native.js#L783-L799), except for a project-specific detail (it returns the stub instead of assigning it to a global map of stubs, since there is no such global map in Positron). Plus, _let_ is the new _var_.

@marcoscaceres How does this look to you?

### DIFF
--- a/positron/modules/positron_util.js
+++ b/positron/modules/positron_util.js
@@ -6,7 +6,7 @@
 
 const { classes: Cc, interfaces: Ci, results: Cr, utils: Cu } = Components;
 
-exports.makeStub = function(signature, returnValue) {
+exports.makeStub = function(signature, { returnValue } = {}) {
   let doNotWarn;
 
   if (typeof returnValue === "function") {
--- a/positron/modules/positron_util.js
+++ b/positron/modules/positron_util.js
@@ -1,0 +1,27 @@
+/* This Source Code Form is subject to the terms of the Mozilla Public
+ * License, v. 2.0. If a copy of the MPL was not distributed with this
+ * file, You can obtain one at http://mozilla.org/MPL/2.0/. */
+
+'use strict';
+
+const { classes: Cc, interfaces: Ci, results: Cr, utils: Cu } = Components;
+
+exports.makeStub = function(signature, returnValue) {
+  let doNotWarn;
+
+  if (typeof returnValue === "function") {
+    doNotWarn = returnValue;
+  } else {
+    doNotWarn = function() { return returnValue };
+  }
+
+  let warnOnce = function() {
+    console.warn(signature + ' not implemented');
+    warnOnce = doNotWarn;
+    return doNotWarn.apply(this, arguments);
+  }
+
+  return function() {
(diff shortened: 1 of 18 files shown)


positron/modules/positron_util.js
```diff
@@ -0,0 +1,27 @@
+/* This Source Code Form is subject to the terms of the Mozilla Public
+ * License, v. 2.0. If a copy of the MPL was not distributed with this
+ * file, You can obtain one at http://mozilla.org/MPL/2.0/. */
+
+'use strict';
+
+const { classes: Cc, interfaces: Ci, results: Cr, utils: Cu } = Components;
+
+exports.makeStub = function(signature, returnValue) {
```
marcoscaceres: Sorry to be a pain, but could we make "returnValue" an object literal instead? The reason being that in various places it looks like a boolean trap. It will also make the API a little more extensible if need be in the future. So:

```js
// optional returnValue, just destructures to undefined when missing
exports.makeStub = function(signature, {returnValue} = {}) {
  // ... other stuff... 
}
thing.foo =  positronUtil.makeStub('process.env', {defaultValue: true});
```

marcoscaceres: fixed code above.

marcoscaceres:
My only concern is the one I commented on... but apart from that, lgtm.

positron/modules/positron_util.js
mykmelez: > Sorry to be a pain, but could we make "returnValue" an object literal instead? The reason being that in various places it looks like a boolean trap. It will also make the API a little more legible + extensible if need be in the future. So:

Ok, I've made the second parameter an object literal with a returnValue property in 00e6c98.

marcoscaceres:
@mykmelez, thanks for fixing those! I'm about to go 💤, as it's 1:30am here. Feel free to merge or I will merge in 10 hours.

Issue #20, opened 2016-04-19
implement <webview> element


mykmelez:
The [&lt;webview> element in Electron](http://electron.atom.io/docs/v0.37.6/api/web-view-tag/) provides an unprivileged frame for loading untrusted web content along with an API for interacting with it, similar to [&lt;iframe mozbrowser> in Gecko](https://developer.mozilla.org/en-US/docs/Mozilla/Gecko/Chrome/API/Browser_API).

It's implemented in Electron by [web-view.js](https://github.com/mozilla/positron/blob/master/positron/electron/lib/renderer/web-view/web-view.js), which calls the native API WebFrame::RegisterEmbedderCustomElement to register the &lt;webview> and &lt;browserplugin> custom elements (as subclasses of HTMLObjectElement) and then exports a WebView() constructor.

The WebView() constructor then creates a &lt;webview> instance, creates a &lt;browserplugin> instance, creates a shadow root on the &lt;webview>, and appends the &lt;browserplugin> to the shadow root.

For Positron, we should be able to enable Web Components, use the standard `document.registerElement` to register a custom &lt;webview> element whose shadow root contains an &lt;iframe mozbrowser>, and then translate &lt;webview> APIs to &lt;iframe mozbrowser> APIs.

We'll probably have to modify web-view.js in this case (although I've been avoiding modifications to the Electron scripts in positron/electron/lib/ generally, since it'll be easier to track upstream changes the less we modify that code).

Also note https://github.com/browserhtml/browserhtml/issues/639, which is about designing a &lt;webview> implementation for Servo/browser.html. Of note, that issue references this [comparison by Kan-Ru](https://wiki.mozilla.org/WebAPI/BrowserAPI/Common_Subset) of &lt;webview> and &lt;iframe mozbrowser>.

mykmelez:
I'm working on this over in https://github.com/mykmelez/positron/tree/webview-element.

mykmelez:
I landed a partial implementation of <webview> in https://github.com/mozilla/positron/pull/60. Unsure if it's worth keeping this issue open for the rest of the implementation. It'd probably make more sense to track individual bits of that implementation in separate issues.

PR #59, opened 2016-05-28
Move electron to a submodule.

brendandahl:
Modules is at v 1.1.3
See fixes in top two commits:
https://github.com/mozilla/positron-electron/commits/master

For now I have electron as the upstream for positron-electron, but we can change that to tofino-electron once I get confirmation that it's being used.


mykmelez:
Nice. This'll complicate certain other changes that require modifications to Electron (like my fix for #20), but it's still useful to be able to track those changes, and the version of Electron we're using, more explicitly.

We should do the same for Node, ideally using the version of Node that Electron itself depends on via a submodule (at least until we replace it with SpiderNode), so we know for sure that we're using a compatible version of Node. But we can do that separately.

The only issue I see is that the README is now out-of-date. It should get updated in conjunction with this change to explain that you need to initialize and update the submodule after cloning (or in conjunction with cloning) before you can build the project.

I also wonder if it makes more sense to put the moz.build instructions into this repo rather than that one. I *think* that's possible, because complete source and destination paths for EXTRA_JS_MODULES can be specified at any level of the tree. And it would be one less change to Electron.

mykmelez:
@brendandahl I saw an issue and have submitted a PR in https://github.com/brendandahl/positron/pull/1 to resolve it.

Now I'm seeing:

> JavaScript error: file:///Users/myk/Projects/positron/positron/test/hello-world/index.html, line 9: NS_ERROR_XPC_JAVASCRIPT_ERROR_WITH_DETAILS: [JavaScript Error: "v8Util.createIDWeakMap is not a function" {file: "resource:///modules/ModuleLoader.jsm -> resource:///modules/renderer/api/remote.js" line: 10}]'[JavaScript Error: "v8Util.createIDWeakMap is not a function" {file: "resource:///modules/ModuleLoader.jsm -> resource:///modules/renderer/api/remote.js" line: 10}]' when calling method: [nsIDOMGlobalPropertyInitializer::init]

I saw that you implemented *createDoubleIDWeakMap* using *WeakMap*. You might be able to do that for *createIDWeakMap* too, but the version of Electron that we're currently using has a different API, an *IDWeakMap* constructor, which I had to implement differently in positron/modules/atom_common_id_weak_map.js in order to get stuff working, since *WeakMap* keys can't be primitives, and the version of Electron we're currently using uses integer primitives as keys (which is presumably also true for the new version, at least for *createIDWeakMap*, although *createDoubleIDWeakMap* looks like it uses objects as keys).

PR #60, opened 2016-05-28
partial implementation of <webview> element

mykmelez:
This is a partial implementation of the &lt;webview> element. It's sufficient to run the webview/browser/ sample app in https://github.com/hokein/electron-sample-apps, and thus it achieves the [M2: Run Basic Browser App](https://github.com/mozilla/positron/wiki/Roadmap#m2-run-basic-browser-app) milestone.


mykmelez:
Ah, hmm, setting the *mozbrowser* and *remote* attributes immediately in the *createdCallback* for the *browserplugin* custom element (3f661b9) resolves that problem, although it then reveals another: an alert dialog appears that says, "This page has a content security policy that prevents it from being loaded in this way." And the URL still doesn't load.

mykmelez:
> Ah, hmm, setting the mozbrowser and remote attributes immediately in the createdCallback for the browserplugin custom element (3f661b9) resolves that problem, although it then reveals another: an alert dialog appears that says, "This page has a content security policy that prevents it from being loaded in this way." And the URL still doesn't load.

Ah, this is triggered by nsDocument::InitCSP, presumably because github.com's CSP doesn't permit framing, and that function thinks it's being loaded in a frame, since there's an *mDocumentContainer* docshell for the *mozbrowser*.

https://dxr.mozilla.org/mozilla-central/rev/4d63dde701b47b8661ab7990f197b6b60e543839/dom/base/nsDocument.cpp#2948

Perhaps we just need to check nsDocShell::GetIsMozBrowserOrApp before deciding whether or not to enforce the frame-ancestor policy, although it returns false in this case, as mFrameType is FRAME_TYPE_REGULAR for the docshell, which seems wrong. So perhaps it's being set incorrectly somewhere.

It could also be an issue that nsCSPContext::PermitsAncestry generates ancestorsArray by walking the docshell tree until it reaches a docshell whose document URI scheme is "chrome", which won't be the case here, as the URL of the document that contains the mozbrowser will be a file: URL like file:///Users/myk/Projects/positron/positron/test/hello-world/index.html.

jryans:
> Perhaps we just need to check nsDocShell::GetIsMozBrowserOrApp before deciding whether or not to enforce the frame-ancestor policy, although it returns false in this case, as mFrameType is FRAME_TYPE_REGULAR for the docshell, which seems wrong. So perhaps it's being set incorrectly somewhere.

You should get `FRAME_TYPE_BROWSER` if `mozbrowser` is set in time and you have permission to use it. It's set by [`nsFrameLoader::MaybeCreateDocShell`][1]. The attribute and permissions check happens in [`nsGenericHTMLFrameElement::GetReallyIsBrowserOrApp`][2]. In desktop Firefox, we have permission implicitly because the embedding page is has the system principal (`nsContentUtils::IsSystemPrincipal` returns true). Perhaps check what `GetReallyIsBrowserOrApp` thinks for your setup?

[1]: https://dxr.mozilla.org/mozilla-central/rev/4d63dde701b47b8661ab7990f197b6b60e543839/dom/base/nsFrameLoader.cpp#2106
[2]: https://dxr.mozilla.org/mozilla-central/rev/4d63dde701b47b8661ab7990f197b6b60e543839/dom/html/nsGenericHTMLFrameElement.cpp#491

mykmelez:
@brendandahl Ok, this is ready for review!

The Electron implementation is complex, but the basic idea is that each &lt;webview> node has a "browser plugin" &lt;object> node in its shadow DOM, a _WebViewImpl_ object in the renderer process, and a _WebContents_ "guest" in the main process.

The "browser plugin" node is the actual rendering frame, the equivalent of an &lt;iframe mozbrowser> in Gecko. And the _WebContents_ guest in the main process is responsible for handling all interaction between the "embedder" _WebContents_ (the _BrowserWindow_ that embeds the &lt;webview> element) and the "guest" _WebContents_, which are supposed to be in separate processes.

There's a complex sequence of initialization steps when a &lt;webview> is created (or parsed from a document being loaded into a _BrowserWindow_): a guest _WebContents_ is created in the main process, then the guest is "attached" to the &lt;webview> (and added to the web view manager, and possibly also the web frame manager—not the same!), and then various other things happen. Oh, and there's a proxy guest in the renderer process that forwards method calls to the actual _WebContents_ guest in the main process via IPC. And, and…

A Positron implementation could be much simpler, since the main and renderer processes are actually the same process, and since the renderer process could access the &lt;iframe mozbrowser> API directly (even after we make it "remote"). But in the interest of maintaining as much Electron compatibility as possible, I minimized changes to the implementation, preferring to support Electron's existing model as much as possible.

The only significant difference is that I made the renderer process pass the _WebViewImpl_ instance back to the main process, so the guest can access the &lt;iframe mozbrowser> directly in order to implement the _WebContents_ API more simply. We could instead make all the _WebContents_ API calls pass messages back to some object in the renderer process that then accesses the &lt;iframe mozbrowser>, but that would be cumbersome and entail even more significant changes to Electron. Let's avoid that for now.

positron/components/Process.js
```diff
@@ -40,6 +40,21 @@ Process.prototype = {
   init: function(window) {
     this._contentWindow = window;
 
+    // The binding appears to get initialized even for a document in a content
```
brendandahl: Now that we have a system principal can't we get rid of this and specify `JavaScript-global-privileged-property` in the manifest instead?

positron/electron/lib/renderer/web-view/web-view.js
```diff
@@ -264,10 +269,92 @@ var WebViewImpl = (function() {
...
+      }
+
+      ipcRenderer.emit("ATOM_SHELL_GUEST_VIEW_INTERNAL_DISPATCH_EVENT-" + internal.viewInstanceId,
+                       event, 'did-stop-loading');
+      // TODO: support the 'did-fail-load' event properties.
+      ipcRenderer.emit("ATOM_SHELL_GUEST_VIEW_INTERNAL_DISPATCH_EVENT-" + internal.viewInstanceId,
+                       event, 'did-fail-load');
+    }, false);
+
+    // this.setAttribute('type', 'application/browser-plugin');
```
brendandahl: Why is this commented out?

positron/components/Process.js
mykmelez: > Now that we have a system principal can't we get rid of this and specify JavaScript-global-privileged-property in the manifest instead?

Indeed! Fixed in 2d5468f.

positron/electron/lib/renderer/web-view/web-view.js
mykmelez: > Why is this commented out?

The _type_ attribute is specific to _object_ elements and isn't defined for _iframe_ elements. I added a comment to that effect and also updated other comments and rearranged code to make it the changes clearer in 688b355.

PR #86, opened 2016-06-14
don't assume the webview has a content document

mykmelez:
The &lt;webview> isn't guaranteed to have a content document, so we should ensure it does before trying to get its URL.


mykmelez:
@brendandahl, does that seem reasonable?

PR #89, opened 2016-06-21
Fix webview focus.

### LINKED ISSUE #82: page loaded in <webview> doesn't respond to key events
After loading a web page in the &lt;webview> element in the [sample browser app](https://github.com/hokein/electron-sample-apps/tree/master/webview/browser), the page doesn't respond to key events such as arrow key presses to scroll the page. Also, clicking on the page doesn't blur the location bar.

If the page has a text field, however, then clicking in that field will focus the field and blur the location bar. After which the page will start responding to key events. The default URL in the sample browser app, http://github.com/, happens to focus a text field on load, so it employs this workaround (albeit inadvertently).

brendandahl:
Upstream bug:
https://bugzilla.mozilla.org/show_bug.cgi?id=1281297

Fixes #82


mykmelez:
Nice.

PR #90, opened 2016-06-22
remove system principal from BrowserWindow document

mykmelez:
@brendandahl This is a work-in-progress shows a bit of what we'd have to do if we removed the system principal from the BrowserWindow document. Obviously I'd like to avoid having to modify Electron code to unwrap the _window_ object every time it wants to touch something on it that the page itself should also see. But I'm not sure how feasible that is.

And there are other consequences, like the way we expose the _Buffer_ global to other modules (although we'll probably have to bind it via WebIDL in the long run anyway). Also, that _process_ global initialization hack doesn't quite work right, I don't think. I see an error in GitHub code with it. Probably because GitHub expects the global to be undefined.

Nevertheless, I suspect that we're better off in the long run if we can get away with not giving the BrowserWindow document the system principal (not only because it fixes &lt;iframe mozbrowser remote>), so I'm going to explore the option more on this branch.


mykmelez:
I'm not actively exploring this option, so I'm going to close this branch for now. I'll reopen if I pick it back up.

Issue #76, opened 2016-06-07
make the mozbrowser iframe remote


mykmelez:
When the &lt;iframe mozbrowser> in the shadow DOM of the &lt;webview> element is remote, it doesn't render its contents. This appears to be a regression from the way we give the BrowserWindow document chrome privileges by assigning it the system principal in nsDocument::reset. We currently work around it by making the iframe not be remote.

That way of giving the BrowserWindow document chrome privileges is probably incorrect, and a fix for that might end up fixing this issue. But I'm filing it here to track it discretely, since it's potentially a separate issue.

We should make the mozbrowser iframe remote because it's more consistent with the behavior of Electron (which creates a renderer process for each webview) and has all the other benefits of e10s wrt. chrome UI responsiveness.

mykmelez:
@jryans Do you have any ideas about the root cause of this by any chance?

jryans:
Hmm, at moment I am not sure. I have actually seen the reverse: for my usage outside of Positron, it only renders when it _is_ remote. How do web components interact with [`nsFrameLoader::Show`][1]? Is your iframe considered remote by the time that runs, or does it run before you had a chance to set it?

[1]: https://dxr.mozilla.org/mozilla-central/rev/3ccccf8e5036179a3178437cabc154b5e04b333d/dom/base/nsFrameLoader.cpp#713

mykmelez:
> Hmm, at moment I am not sure. I have actually seen the reverse: for my usage outside of Positron, it only renders when it is remote. How do web components interact with nsFrameLoader::Show? Is your iframe considered remote by the time that runs, or does it run before you had a chance to set it?

_nsFrameLoader::Show_ is called and calls _IsRemoteFrame_ twice during startup of the [basic browser app](https://github.com/hokein/electron-sample-apps/tree/master/webview/browser), and if the &lt;webview> implementation is setting the 'remote' attribute (i.e. https://github.com/mozilla/positron/blob/639915958d8dc8a51fd5e5f1133bd3fb1c0df962/positron/electron/lib/renderer/web-view/web-view.js#L296-L298 is uncommented), then _IsRemoteFrame_ returns `true` the first time and `false` the second time.

I also see the message "Content sandbox disabled due to sandbox level setting" in the terminal. And I get a download dialog for a file that turns out to be the HTML document at the URL that the &lt;webview> is loading.

mykmelez:
[...]

PR #91, opened 2016-06-24
Ensure remote iframes don't get the system principal.

brendandahl:
After some careful log review of remote=true/false, I saw that there were some NS_FAILED that'd didn't occur on remote=false. After checking those out, I tracked the failure down to https://github.com/mozilla/gecko-dev/blob/965dbc5745f2dd40106a3e075ab8d3ef284c0aa6/caps/nsSystemPrincipal.cpp#L73

Turns out our hack to give system principal to the top document was also giving it to the remote iframe.


brendandahl:
Fixes #76

mykmelez:
Yikes, nice catch!

PR #95, opened 2016-06-24
Track the url of remote iframes.

### LINKED ISSUE #92: GuestWebContentsPrototype.getURL fails because contentDocument is null
@brendandahl, #91 appears to have regressed _GuestWebContentsPrototype.getURL_, as _this._webView.browserPluginNode.contentDocument_ is `null`. So the URL bar in the sample browser becomes empty after loading a page. Which makes sense, since the document is now in the remote process, so it can't be dereferenced in the local one.

Perhaps the solution here is to maintain the current value of the URL locally, updating it each time it changes via the mozbrowserlocationchange event handler.

brendandahl:
The sample browser app still leaves a bit to be desired, but it seems to also have issues with electron as well (I see "undefined" flashed in the url bar when loading and changing pages).

Fixes #92

### DIFF
--- a/positron/modules/atom_browser_web_contents.js
+++ b/positron/modules/atom_browser_web_contents.js
@@ -128,18 +128,20 @@ const BrowserWindowWebContentsPrototype = {
 
 const GuestWebContentsPrototype = {
   _webView: null,
+  _url: null,
   attachWebViewToGuest(webView) {
     this._webView = webView;
+
+    let onBrowserLocationChange = (e) => {
+      this._url = e.detail;
+    };
+    this._webView.browserPluginNode.addEventListener("mozbrowserlocationchange", onBrowserLocationChange);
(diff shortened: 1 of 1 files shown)


mykmelez:
> The sample browser app still leaves a bit to be desired, but it seems to also have issues with electron as well (I see "undefined" flashed in the url bar when loading and changing pages).

Indeed. Another issue I've noticed is that the page width is greater than the window width, so a horizontal scrollbar appears, which you can use to scroll the &lt;webview> slightly out of view. That happens in both Positron and Electron.

It might be worth contributing changes to that app to make it a better demo, since it's currently our demo app for Positron. We could even import it into our repository (or make it a submodule, perhaps forking it first so we can iterate more quickly than upstream). Provided we don't specialize it for Positron unintentionally.

PR #96, opened 2016-06-29
merge from upstream

mykmelez:
@jryans After [bug 1281731](https://bugzilla.mozilla.org/show_bug.cgi?id=1281731) landed, I merged from upstream and resolved conflicts in toolbox-init.js. I think I did so correctly, and the toolbox opens as expected, but when I click another tab in it, the toolbox closes.

If I'm in the Console tab and click the Debugger tab, then it simply closes without printing anything to the terminal. But if I'm in the Debugger tab and click the Console tab, it closes and prints this message: `JavaScript error: resource://gre/modules/commonjs/toolkit/loader.js -> resource://devtools/client/framework/toolbox.js, line 285: TypeError: this._host is null`.

(The toolbox reopens to the last tab I clicked, which is how I can start in different tabs. Clicking on other tabs also exhibit the same behavior of causing the toolbox to close, with varying output to the terminal.)

(The remote toolbox for debugging Positron itself works correctly. It's only the in-process toolbox for debugging the application that has this problem.)


jryans:
Okay, I'll take a look!

mykmelez:
Merging with @jryans's fix ff6bacd for the toolbox issue.

Issue #45, opened 2016-05-19
figure out how to make cross-sandbox Object.prototype comparisons


mykmelez:
https://github.com/mozilla/positron/blob/master/positron/electron/lib/browser/rpc-server.js#L46 checks if a prototype object is strictly equal to *Object.prototype*, i.e. `proto === Object.prototype`. But the prototype can come from a different sandbox than the one in which the comparison is taking place, in which case the condition evaluates to false, even if *proto* is an *Object.prototype*, perhaps because each sandbox gets its own unique instance of *Object*, so it isn't the same *Object.protoype*.

In #26, I work around the problem with a duck-type comparison. But that requires changing the code, which comes from upstream Electron. Ideally, we wouldn't have to modify the Electron code in order to ensure that conditionals like this one behave as it expects.

mykmelez:
@khuey do you have insight here into how sandboxes behave, and how we might configure them to have the behavior we desire?

Note that there isn't a security concern here. Both sandboxes are fully privileged and are loading modules that belong to the app itself. So giving them the same *Object* global wouldn't introduce a vulnerability. And that may be what Node does (unsure).

mykmelez:
> Note that there isn't a security concern here. Both sandboxes are fully privileged and are loading modules that belong to the app itself. So giving them the same Object global wouldn't introduce a vulnerability. And that may be what Node does (unsure).

Or rather: I don't think there's a security concern. But I'm happy to be disabused of my naiveté. ;-)

khuey:
I don't know much about sandboxes ... that's more of a @bholley question.

bholley:
Can you clarify a bit more what's going on here? Does Eletron run code in one global Positron we run it in two separate ones?

Separate prototypes are not unique to sandboxes and are totally something that happens on the web (i.e. same-origin iframes).

mykmelez:
> Can you clarify a bit more what's going on here?

Sure!

> Does Eletron run code in one global Positron we run it in two separate ones?

Electron uses Node to load modules, and Node loads them in a way that shares globals (or at least appears to do so). For example, given a Node app with two modules:

foo.js:
```js
let bar = require("./bar"); // import the "bar" object from bar.js
// Is bar's prototype identical to the Object.prototype in this module?
console.log(Object.getPrototypeOf(bar) === Object.prototype);
```

bar.js:
```js
module.exports = {};
```

[...]

PR #97, opened 2016-06-29
Load modules in their own scopes with evalInSandbox.

### LINKED ISSUE #93: loading renderer modules in single sandbox throws "TypeError: 'get location' called on an object that does not implement interface Window."
In the renderer process, we can't yet load all modules in a single sandbox, because attempts to reference Window properties on the target object then throw errors like: `TypeError: 'get location' called on an object that does not implement interface Window.`

This is described as "TypeError: invalid Array.prototype.sort argument" in the [MDN docs on the error](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Errors/Array_sort_argument). But that's clearly not the issue here.  Some code is deciding that the target object doesn't implement Window, even though its prototype is a Window instance.

Whereas that code does think that a sandbox with a Window prototype implements Window.  So we work around the problem by loading modules in their own sandboxes in renderer processes. But we should resolve the underlying issue so we can load them in the same sandbox to avoid issues like #45, which we worked around in #26.

(That issue happens to be specific to the main process. But similar issues could crop up in the renderer process.)

brendandahl:
Fixes #93 an #45

This solution doesn't seem quite as clean as using the built in loadSubScript and I  wonder about the performance implications of loading the entire script into a string and evaling it.

We could instead try to fix the DOM bindings issue I mentioned in https://github.com/mozilla/positron/issues/45#issuecomment-228914035

Thoughts?

### DIFF
--- a/positron/components/ModuleLoader.jsm
+++ b/positron/components/ModuleLoader.jsm
@@ -14,6 +14,7 @@ const Cr = Components.results;
 const Cu = Components.utils;
 
 Cu.import('resource://gre/modules/Services.jsm');
+Cu.import("resource://gre/modules/NetUtil.jsm");
 
 const subScriptLoader = Cc['@mozilla.org/moz/jssubscript-loader;1'].
                         getService(Ci.mozIJSSubScriptLoader);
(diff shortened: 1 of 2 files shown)


mykmelez:
> This solution doesn't seem quite as clean as using the built in loadSubScript and I wonder about the performance implications of loading the entire script into a string and evaling it.

Indeed! However, _mozIJSSubScriptLoader.loadSubScript_ with a _targetObj_ also has perf implications.



> We could instead try to fix the DOM bindings issue I mentioned in #45 (comment)

Let's do both: land this to fix those issues in the short term, then figure out a fix for the DOM bindings issue and switch back to _loadSubScript_ in conjunction with it.

mykmelez:
> Let's do both: land this to fix those issues in the short term, then figure out a fix for the DOM bindings issue and switch back to loadSubScript in conjunction with it.

I merged this, which closed #93, but it left #45 open. We can use #45 to track the DOM bindings issue or close it as well and file another issue for that, either way is fine with me.

Issue #98, opened 2016-06-29
Find or implement an event similar to blink's NavigationEntryCommitted


brendandahl:
In atom_browser_web_contents.js we need to emit an event with more information so NavigationController tell if the location change was inPage and if it is suppose to replaceEntry.

PR #99, opened 2016-06-29
Emit a 'navigation-entry-commited' event so NavigationController upda…

brendandahl:
…tes the url.


brendandahl:
Follow up issue #98

PR #106, opened 2016-08-04
Move electron to a submodule.

brendandahl:
To get things going I'm going back to 6041c7edf9fc8ff4d73b0c311200c3953127da3e which is where we copied electron/libs

I was going to rebase to remove the reverts, but git seems to have issues with rebasing and having a submodule added that is the same name as an existing folder. Maybe we can just do a squash?


positron/modules/atom_common_v8_util.js
```diff
@@ -11,3 +11,8 @@ exports.getHiddenValue = util.getHiddenValue;
 exports.setHiddenValue = util.setHiddenValue;
 
 exports.setDestructor = positronUtil.makeStub('atom_common_v8_util.setDestructor');
+
+exports.createDoubleIDWeakMap = function() {
+  // TODO: Properly implment this.
+  return new WeakMap();
+};
```
mykmelez: Do we actually need this yet? I think this is only needed once we upgrade Electron.

positron/moz.build
```diff
@@ -2,4 +2,93 @@
 # License, v. 2.0. If a copy of the MPL was not distributed with this
 # file, You can obtain one at http://mozilla.org/MPL/2.0/.
 
-DIRS += ['app', 'chrome', 'components', 'electron', 'modules', 'node', 'webidl']
+DIRS += ['app', 'chrome', 'components', 'modules', 'node', 'webidl']
+
+EXTRA_JS_MODULES.browser += [
```
mykmelez: It's unfortunate that we have to list each module explicitly, especially because we'll have to touch this on Electron updates. But I suppose it has the advantage that we can choose to build only a subset of modules in the future.

PR #111, opened 2016-08-17
enable mozbrowser API for specific origins

mykmelez:
@brendandahl This branch enables mozbrowser for specific origins instead of giving the system principal to all top-level docshell loads, per comments from @bz and @jryans in [bug 1290272](https://bugzilla.mozilla.org/show_bug.cgi?id=1290272).

In theory, it should enable us to expose mozbrowser to BrowserWindow instances without giving them the system principal. But in practice, some of the code we load in BrowserWindow instances needs that principal in order to access Components, so we'd need to disentangle that code from the code that doesn't. So this isn't ready to land yet.

### DIFF
--- a/dom/webidl/BrowserElement.webidl
+++ b/dom/webidl/BrowserElement.webidl
@@ -30,32 +30,32 @@ BrowserElement implements BrowserElementPrivileged;
 interface BrowserElementCommon {
   [Throws,
    Pref="dom.mozBrowserFramesEnabled",
-   ChromeOnly]
+   Func="nsDocument::IsBrowserElementEnabled"]
   void setVisible(boolean visible);
 
   [Throws,
    Pref="dom.mozBrowserFramesEnabled",
-   ChromeOnly]
+   Func="nsDocument::IsBrowserElementEnabled"]
   DOMRequest getVisible();
 
   [Throws,
    Pref="dom.mozBrowserFramesEnabled",
-   ChromeOnly]
+   Func="nsDocument::IsBrowserElementEnabled"]
   void setActive(boolean active);
 
   [Throws,
    Pref="dom.mozBrowserFramesEnabled",
-   ChromeOnly]
+   Func="nsDocument::IsBrowserElementEnabled"]
   boolean getActive();
 
   [Throws,
    Pref="dom.mozBrowserFramesEnabled",
-   ChromeOnly]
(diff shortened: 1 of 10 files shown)


dom/webidl/BrowserElement.webidl
```diff
@@ -30,40 +30,40 @@ BrowserElement implements BrowserElementPrivileged;
 interface BrowserElementCommon {
   [Throws,
    Pref="dom.mozBrowserFramesEnabled",
```
jryans: This pref is also tested in `AudioChannelManager.webidl`, `BrowserElementAudioChannel.webidl`, and `BrowserElementProxy.webidl`. Not sure if you care about those cases though.

mykmelez: > This pref is also tested in AudioChannelManager.webidl, BrowserElementAudioChannel.webidl, and BrowserElementProxy.webidl. Not sure if you care about those cases though.

@jryans: thanks for the tip! I don't need those for this use case, but it may make sense to be consistent.

> In theory, it should enable us to expose mozbrowser to BrowserWindow instances without giving them the system principal. But in practice, some of the code we load in BrowserWindow instances needs that principal in order to access Components, so we'd need to disentangle that code from the code that doesn't.

@brendandahl More specifically: there are modules that need to access _Components_, like the binding modules. And then there are modules that need to expose APIs to the web page, like the `<webview>` implementation, which exposes a _WebView_ constructor, some methods that it exposes [via the constructor's prototype](https://github.com/mozilla/positron-electron/blob/master/lib/renderer/web-view/web-view.js#L433-L435), and some attributes that it [sets on each WebView instance](https://github.com/mozilla/positron-electron/blob/master/lib/renderer/web-view/web-view-attributes.js#L46-L55).

The former modules just need to be loaded in a context that gives them access to _Components_, perhaps the current module sandbox with the system principal. The latter modules are more complicated, however.

They can't be loaded in a context with the system principal as they're currently implemented, since they assume that the page has transparent access to the methods and attributes they define, which xray wrappers prevent across origins for script that doesn't have the system principal.

We could rewrite those modules to expose the APIs via WebIDL or the like, but that's likely to be a significant amount of work. We might also be able to load them in a context with the same principal as the web page, perhaps the page itself, or a sandbox with its principal. Unsure about all the implications here.

mykmelez:
Work on this has stalled, so I'm going to close the pull request for now.

PR #113, opened 2016-08-23
[WIP] Fix the libuv event loop for mac and linux.

brendandahl:
Event loop now runs and it's possible to do a hello world example with a server, but I'm currently getting a compartment mismatch crash on shutdown. I also have a have few TODO's in there that I'm still looking at to see if I need to address them.

example changes to hello world

```(diff)
+++ b/positron/test/hello-world/main.js
@@ -24,7 +24,7 @@ app.on('ready', function() {
     mainWindow = new BrowserWindow({width: 800, height: 600});
 
     // and load the index.html of the app.
-    mainWindow.loadURL('file://' + __dirname + '/index.html');
+    mainWindow.loadURL('http://localhost:8000');
 
     // Open the DevTools.
     mainWindow.webContents.openDevTools();
@@ -37,3 +37,15 @@ app.on('ready', function() {
        mainWindow = null;
     });
 });
+
+// Load the http module to create an http server.
+var http = require('http');
+
+// Configure our HTTP server to respond with Hello World to all requests.
+var server = http.createServer(function (request, response) {
+  response.writeHead(200, {"Content-Type": "text/plain"});
+  response.end("Hello World from node " + process.versions.node + "\n");
+});
+
+// Listen on port 8000, IP defaults to 127.0.0.1
+server.listen(8000);
```

### DIFF
--- a/positron/components/PositronAppRunner.jsm
+++ b/positron/components/PositronAppRunner.jsm
@@ -63,7 +63,7 @@ PositronAppRunner.prototype = {
     };
     let nodeLoader = Cc["@mozilla.org/positron/nodeloader;1"]
                      .getService(Ci.nsINodeLoader);
-    nodeLoader.init();
+    nodeLoader.init("browser");
   },
 
   _parsePackageJSON: function par_parsePackageJSON(aData) {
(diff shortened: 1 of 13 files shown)


mykmelez:
FWIW, I'm not seeing that crash on shutdown over in #119.

PR #116, opened 2016-09-07
merge upstream into node-integration branch

mykmelez:
@brendandahl This merges to the same upstream commit as current SpiderNode and updates the SpiderNode integration to use @ehsan's new _--with-external-spidermonkey-*_ flags (with the fixes in https://github.com/mozilla/spidernode/pull/193 to get SpiderNode to build with an external SpiderMonkey; and with some modifications to make those flags work for this use case).


mykmelez:
Note: the changes I made here are fairly minimal, just https://github.com/mozilla/positron/pull/116/commits/481e47dc6cb3ec0afa11781471dd98cafeac2dbd.

I suspect those could also use some improvement. I'm not too happy about the order in which we build and link Positron/SpiderMonkey and SpiderNode. It feels brittle and buggy. I'm still thinking on how to improve that. In the meantime, however, this should at least unblock Positron development on this branch.

mykmelez:
> I suspect those could also use some improvement. I'm not too happy about the order in which we build and link Positron/SpiderMonkey and SpiderNode. It feels brittle and buggy. I'm still thinking on how to improve that. In the meantime, however, this should at least unblock Positron development on this branch.

I've still been seeing occasional build failures that SpiderNode doesn't have a rule to make libjs_static.a, which suggests that SpiderNode is sometimes being built before SpiderMonkey.

The best way to ensure that SpiderMonkey always gets built first is to integrate the build systems, generating moz.build files from SpiderNode GYP files using MozBuild's GYP reader (with enhancements). But that's a larger project, and in the meantime, fbd225e may help.

PR #118, opened 2016-09-19
partially support Tofino

mykmelez:
@brendandahl This lands some fixes to improve support for Tofino on the node-integration branch.

It doesn't fix all issues, as there is still an error being thrown in Tofino's main process, and the browser process fails because it still requires Node there. But it does enable Tofino to open a BrowserWindow. It just doesn't yet load the browser's UI into it.

Some changes to Tofino were also required, and I've landed those in the support-positron branch of my Tofino fork (https://github.com/mykmelez/tofino/tree/support-positron). Most of the changes are just workarounds for `let (const of)`, which we can undo when we merge from upstream gecko-dev, where that's been implemented. Another change stubs out the use of a Node logging module that depends on a native module.

This branch includes your libuv event loop fixes in 867fb9c26dbd1f731a2f6aae5bebcf98ca42ecab. FWIW, on my Mac, all of the test apps (hello-world, hello-world-server, and electron-sample-apps/webview/browser/) work as expected, and none of them crash on shutdown (in my optimized build).


mykmelez:
Erm, I requested that you pull this into master, when I meant to request that you pull it into node-integration. I don't see a way to change the target branch, so I'll close this and open a new pull request.

PR #119, opened 2016-09-19
partially support tofino

mykmelez:
@brendandahl This lands some fixes (mostly stubs) to improve support for Tofino on the node-integration branch.

It doesn't fix all issues, as there is still an error being thrown in Tofino's main process, and the browser process fails because it still requires Node there. But it does enable Tofino to open a BrowserWindow. It just doesn't yet load the browser's UI into it.

Some changes to Tofino were also required, and I've landed those in the support-positron branch of my Tofino fork (https://github.com/mykmelez/tofino/tree/support-positron). Most of the changes are just workarounds for `let (const of)`, which we can undo when we merge from upstream gecko-dev, where that's been implemented. Another change stubs out the use of a Node logging module that depends on a native module.

This branch includes your libuv event loop fixes in 867fb9c26dbd1f731a2f6aae5bebcf98ca42ecab. FWIW, on my Mac, all of the test apps (hello-world, hello-world-server, and electron-sample-apps/webview/browser/) work as expected, and none of them crash on shutdown (in my optimized build).

### DIFF
--- a/positron/components/PositronAppRunner.jsm
+++ b/positron/components/PositronAppRunner.jsm
@@ -63,7 +63,7 @@ PositronAppRunner.prototype = {
     };
     let nodeLoader = Cc["@mozilla.org/positron/nodeloader;1"]
                      .getService(Ci.nsINodeLoader);
-    nodeLoader.init();
+    nodeLoader.init("browser");
   },
 
   _parsePackageJSON: function par_parsePackageJSON(aData) {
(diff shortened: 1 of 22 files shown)


mykmelez:
(This also includes #114.)

brendandahl:
> in my optimized build)

The crashes are assertions, so I'd expect it's fine in the optimized builds.

PR #121, opened 2016-09-20
taskcluster improvements

mykmelez:
This does a variety of things, in particular:

* build debug in addition to optimized (although it still doesn't test either build);
* simplify and clarify task commands (mostly formatting, but a few other things);
* speed up tasks by disabling auto `git gc` (which can take many minutes and isn't necessary in automation).

We should eventually disable auto `git gc` in the Docker image, but this'll do for now.

We should probably squash this branch, since it has a bunch of small changes that aren't particularly interesting of themselves.


mykmelez:
> TaskCluster-GitHub: https://tools.taskcluster.net/task-graph-inspector/#NFbkLaNiS_CVE01iwpdQrA

This is *not* actually testing the commit it says it's testing, because TaskCluster has decided to use the original .taskcluster.yml file, not the one from the commit, for some reason. Even stranger, it actually triggered two task graph runs, the one in the TaskClusterRobot comment and this one in the status "dot" link (and Details link): https://tools.taskcluster.net/task-graph-inspector/#N-Pj3JQgRQeSu0Nd-xF6PA (both of which are using the original .taskcluster.yml file).

I filed https://github.com/taskcluster/taskcluster-github/issues/72 on the TaskCluster strangeness. In the meantime, I think this is safe to merge, since the task graph run on the second-to-last commit looked like it was going to succeed, if only it didn't time out: https://tools.taskcluster.net/task-graph-inspector/#RCnYNzxwQXyNZuYRaYOBGg , and the last commit just increases the timeout from 2 hours to 3 hours.

PR #125, opened 2016-09-21
Node integration

brendandahl:


.gitmodules
```diff
@@ -1,3 +1,7 @@
 [submodule "positron/electron"]
 	path = positron/electron
 	url = https://github.com/mozilla/positron-electron.git
+[submodule "positron/spidernode"]
+	path = positron/spidernode
+	url = https://github.com/mozilla/positron-spidernode.git
+	branch = existing-js-runtime
```
mykmelez: This branch reference is obsolete, but I'm unsure it actually matters, since it doesn't seem to affect the result if `git submodule update`.

PR #142, opened 2016-10-04
force-copy SpiderNode libs

mykmelez:
A TaskCluster build failed because these were already present. I'm unsure why, but perhaps TaskCluster had cached and was reusing the Positron working directory. The Makefile should force-copy the libs regardless.


mykmelez:
One of the two jobs in #146 failed because of this issue. Landing as a bustage fix.

PR #148, opened 2016-10-12
Update spidernode to fix leaks.

brendandahl:


mykmelez:
Looks good, and works locally. I'll just wait for TaskCluster to return results before merging.

mykmelez:
The opt build succeeded, but the debug build failed because of unrelated issue #149.

PR #151, opened 2016-10-30
Gecko dev master

AliaTabassumAS:


mykmelez:
Hi @AliaTabassumAS, thanks for the pull request! In order to update Positron to a newer version of Gecko, we need to also update https://github.com/mozilla/positron-spidernode to a version of SpiderNode that has been synced to the same upstream version of Gecko.

And unfortunately that's complicated, since SpiderNode uses https://github.com/glandium/git-cinnabar to sync with the Mercurial repository for Gecko, while Positron uses https://github.com/mozilla/gecko-dev, and the two have different revision histories.

In any case, I notice that you've recently opened a variety of pull requests for different repositories, none of which have included a description, and all of which appear to be requests to pull some branch in a repository into that repository's master branch. Some of those requests have already been closed, and in a couple cases, the closer commented that the request didn't make sense.

So it looks like this request is similarly made without cause, and I'm going to close it. Please reopen if you really do want to see this happen and can explain your reasoning!

PR #153, opened 2016-12-26
merge upstream Gecko again

mykmelez:
I didn't notice this when I merged #152, but for some reason the Gecko merge produced a non-merge commit containing the diff between the two branches instead of a merge commit that add the upstream commits as a branch.

I have no idea why, since I didn't do anything unusual, just `git merge gecko-dev/master` (where "gecko-dev" is the name of the remote for https://github.com/mozilla/gecko-dev).

In any case, here's a merge commit that actually merges the branches. The command I invoked is the same, `git merge gecko-dev/master`, but the output is what I expected the first time around (and should have confirmed before merging the pull request).

On this branch, when I `git merge gecko-dev/master`, then Git says the branch is "Already up-to-date." So merging this branch should get things into a good state.

### DIFF
--- a/positron/config/mozconfig
+++ b/positron/config/mozconfig
@@ -1,4 +1,3 @@
 ac_add_options --disable-export-js
 ac_add_options --disable-shared-js
 ac_add_options --disable-js-shell
-ac_add_options --enable-sm-promise


mykmelez:
> On this branch, when I git merge gecko-dev/master, then Git says the branch is "Already up-to-date." So merging this branch should get things into a good state.

So optimistic! In fact there are (at least) two additional issues:
* The [docker-build-positron image](https://hub.docker.com/r/mykmelez/docker-build-positron/) is out-of-date, as Gecko now requires GCC 4.9+ (the image has 4.8) and Rust (which bootstrap.py will install but won't ensure is on the PATH). I've updated the image from Trusty to Xenial, which gets us a newer GCC; and I've added an ENV command to the Dockerfile that ensures Rust executables are on the PATH.
* The positron/config/mozconfig file sets the now-obsolete configure option `--enable-sm-promise`. I've removed that in a091040.

mykmelez:
Next issue: the taskcluster-vcs Node module was failing due to https://github.com/taskcluster/taskcluster-vcs/issues/68, so I added a workaround to docker-build-positron. The workaround doesn't require any changes to Positron, but it's hard to kick TaskCluster without a commit, so I did it via the `--allow-empty` commit 6a38657.

PR #155, opened 2017-07-24
Ensure that path exists

Dyex719:
This pull request is in response to Bug 1369710 - [mozlint] Error if unrecognized path is passed in.
More info: https://bugzilla.mozilla.org/show_bug.cgi?id=1369710


mykmelez:
This change isn't specific to Positron, so it isn't something we can merge into this repository. Over in bug 1369710, I've made some recommendations for how to proceed with this change in the upstream repository.

